PR #6, opened 2024-02-13
Initial command and class skeleton

happz:


newa/__init__.py
```diff
@@ -1,6 +1,132 @@
...
+
+
+T = TypeVar('T')
+SerializableT = TypeVar('SerializableT', bound='Serializable')
+
+
+def yaml_parser() -> ruamel.yaml.YAML:
+    """ Create standardized YAML parser """
+
+    yaml = ruamel.yaml.YAML()
```
lukaszachy, line 25: Should we play safe?
```suggestion
    yaml = ruamel.yaml.YAML(typ='safe')
```

happz: Fixed in 4d34e7f

kkaarreell:
I think the `event` processor should be using different filename prefix for input (containing `[event]`) and produced files (containing also `[errata]`).

kkaarreell (approved):
The current version LGTM.  Thank you.

kkaarreell:
> I think the `event` processor should be using different filename prefix for input (containing `[event]`) and produced files (containing also `[errata]`).



> I think the `event` processor should be using different filename prefix for input (containing `[event]`) and produced files (containing also `[errata]`).



> I think the `event` processor should be using different filename prefix for input (containing `[event]`) and produced files (containing also `[errata]`).

This has been addressed by introducing`init-` prefix for input files.

thrix (approved):
LGTM

PR #7, opened 2024-02-15
Add skeleton of issue actions

happz:


newa/cli.py
```diff
@@ -94,14 +95,42 @@ def cmd_jira(ctx: CLIContext) -> None:
...
+
+        while issue_actions:
+            action = issue_actions.pop(0)
+
+            print(f'* Would create a {action.type.name} issue:')
+            print(f'     summary: {action.summary}')
+            print(f'     summary: {action.description}')
+
+            if action.id in known_issues:
+                raise Exception(f'Issue "{action.id}" is already created!')
```
jchecahi, line 116: Does it make sense to raise an Exception here? In a respin, the epic would already exist, but `newa` would still need to create a new task (and subsequent subtasks) parented to it. Assuming that created/existing issues were queried just before (as mentioned in [line 100](https://github.com/RedHatQE/newa/pull/7/files#diff-fbfe615aa583117b7a57302f5da6ff45c1eda26a6f4b2406b68a09ffe1bd698aR100)), wouldn't it be more suitable to log the message and start another loop iteration?

happz: Good point, and it's definitely up to you to replace it with more fitting handling. As it is now, it exists only to do something and fail if there's a suspicious issue. I fully expect the whole loop to look completely different in a couple of weeks, once Jira-related code starts landing in :)

kkaarreell (approved):
I tested this and it works OK. Thank you.

kkaarreell:
@thrix could you please merge this one? I don't seem to have permissions to do so.

thrix (approved):
LGTM

jchecahi:
LGTM!

PR #12, opened 2024-03-04
Implement initial data fetch from errata

### LINKED ISSUE #10: Implement fetch of errata data from Errata Tool
We need to implement fetch of errata data from Errata Tool. The idea is to load all necessary data from errata tool for a given single errata. We are not 100% sure what data will be exactly needed but it should be easy to extend or narrow the scope of it. The logic should most likely belong to fetch_details method of Erratum class. It might or might not be necessary to add an accompanying new class to hold the data. Nevertheless. for the start it will be okay to keep the data as Erratum class attributes.

The-Mule:
Initial implementation of fetching data from errata tool. You need to have kerberos ticket to access errata tool, you need to provide CA bundle (`REQUESTS_CA_BUNDLE`) and a link to ET (`NEWA_ET_URL`).

Fixes #10

### DIFF
--- a/newa/__init__.py
+++ b/newa/__init__.py
@@ -160,7 +160,7 @@ class ErrataTool:
     # url via environment variable NEWA_ET_URL
     url: str = field(validator=validators.matches_re("^https?://.+$"))
 
-    @url.default # pyright: ignore
+    @url.default  # pyright: ignore [reportAttributeAccessIssue]
     def _url_factory(self) -> str:
         if "NEWA_ET_URL" in os.environ:
             return os.environ["NEWA_ET_URL"]
--- a/newa/__init__.py
+++ b/newa/__init__.py
@@ -1,14 +1,20 @@
+from __future__ import annotations
+
 import io
+import os
+import time
 from enum import Enum
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Optional, TypeVar
 
 import attrs
 import jinja2
+import requests
 import ruamel.yaml
 import ruamel.yaml.nodes
 import ruamel.yaml.representer
-from attrs import define, field
+from attrs import define, field, frozen, validators
+from requests_kerberos import HTTPKerberosAuth
 
 if TYPE_CHECKING:
     from typing_extensions import Self, TypeAlias
@@ -82,6 +88,19 @@ def render_template(
         raise Exception("Could not render template.") from exc
 
 
+def krb_get_request(url: str, attempts: int = 5, delay: int = 5) -> Any:
+    """ Generic GET request using Kerberos authentication """
+
+    while attempts:
+        r = requests.get(url, auth=HTTPKerberosAuth(delegate=True))
+        if r.status_code == 200:
+            return r.json()
+        time.sleep(delay)
+        attempts -= 1
+
+    raise Exception(f"GET request to {url} failed")
+
+
 class EventType(Enum):
     """ Event types """
 
@@ -92,7 +111,7 @@ class EventType(Enum):
 class Cloneable:
     """ A class whose instances can be cloned """
 
-    def clone(self) -> 'Self':
+    def clone(self) -> Self:
         return attrs.evolve(self)
 
 
@@ -126,7 +145,33 @@ class Event(Serializable):
     """ A triggering event of Newa pipeline """
 
     type_: EventType = field(converter=EventType)
-    id: 'ErratumId'
+    id: ErratumId
+
+
+@frozen
+class ErrataTool:
+    """
+    Interface to Errata Tool instance
+
+    Its only purpose is to serve information about an erratum and its builds.
+    """
+
+    # TODO: Until we have a dedicated newa config, we'll setup ET instance
+    # url via environment variable NEWA_ET_URL
+    url: str = field(validator=validators.matches_re("^https?://.+$"))
+
+    @url.default # pyright: ignore
+    def _url_factory(self) -> str:
+        if "NEWA_ET_URL" in os.environ:
+            return os.environ["NEWA_ET_URL"]
+        raise Exception("NEWA_ET_URL envvar is required.")
+
+    # TODO: Not used at this point because we only consume builds now
+    def fetch_info(self, erratum_id: str) -> Any:
+        return krb_get_request(f"{self.url}/advisory/{erratum_id}.json")
+
+    def fetch_releases(self, erratum_id: str) -> Any:
+        return krb_get_request(f"{self.url}/advisory/{erratum_id}/builds.json")
 
 
 @define
@@ -145,13 +190,63 @@ class InitialErratum(Serializable):
 
 @define
 class Erratum(Cloneable, Serializable):
-    """ An eratum """
+    """
+    An eratum
 
+    Represents a set of builds targetting a single release.
+    """
+
+    # TODO: We might need to add more in the future.
     release: str
     builds: list[str] = field(factory=list)
 
-    def fetch_details(self) -> None:
-        raise NotImplementedError
+    @classmethod
+    def from_errata_tool(cls, event: Event) -> list[Erratum]:
+        """
+        Creates a list of Erratum instances based on given errata ID
+
+        Errata is split into one or more instances of an erratum. There is one
+        for each release included in errata. Each errata has a single release
+        set - it is either regular one or ASYNC. An errata with a regular
+        release (e.g. RHEL-9.0.0.Z.EUS) will result into a single erratatum.
+        On the other hand an errata with ASYNC release might result into one
+        or more instances of erratum.
+        """
+
+        errata = []
+
+        # TODO: We might need to assert QE (or later) state. There is no point
+        # in fetching errata in NEW_FILES where builds need not to be present.
+
+        # In QE state there is are zero or more builds in an erratum, each
+        # contains one or more packages, e.g.:
+        # {
+        #   "RHEL-9.0.0.Z.EUS": [
+        #     {
+        #       "scap-security-guide-0.1.72-1.el9_3": {
+        #          ...
+        #     }
+        #   ]
+        #   "RHEL-9.2.0.Z.EUS": [
+        #     {
+        #       "scap-security-guide-0.1.72-1.el9_3": {
+        #          ...
+        #     }
+        #   ]
+        # }
+
+        releases_json = ErrataTool().fetch_releases(event.id)
+        for release in releases_json:
+            builds = []
+            builds_json = releases_json[release]
+            for item in builds_json:
+                builds += list(item.keys())
+            if builds:
+                errata.append(cls(release, builds))
+            else:
+                raise Exception(f"No builds found in ER#{event.id}")
+
+        return errata
 
 
 @define
@@ -169,9 +264,6 @@ class Recipe(Cloneable, Serializable):
     """ A job recipe """
 
     url: str
-
-    def fetch_details(self) -> None:
-        raise NotImplementedError
 
 
 @define
--- a/newa/cli.py
+++ b/newa/cli.py
@@ -112,30 +112,27 @@ def main(click_context: click.Context, state_dir: str) -> None:
     multiple=True,
     )
 @click.pass_obj
-def cmd_event(ctx: CLIContext, errata_ids: tuple[str, ...]) -> None:
+def cmd_event(ctx: CLIContext, errata_ids: list[str]) -> None:
     ctx.enter_command('event')
 
-    if errata_ids:
-        for erratum_id in errata_ids:
-            event = Event(type_=EventType.ERRATUM, id=erratum_id)
+    # Errata IDs were not given, try to load them from init- files.
+    if not errata_ids:
+        errata_ids = [e.event.id for e in ctx.load_initial_errata('init-')]
 
-            # fetch erratum details, namely releases
-            releases = ['RHEL-8.10.0', 'RHEL-9.4.0']
+    # Abort if there are still no errata IDs.
+    if not errata_ids:
+        raise Exception('Missing errata IDs!')
 
-            for release in releases:
-                erratum_job = ErratumJob(event=event, erratum=Erratum(release=release))
+    for erratum_id in errata_ids:
+        event = Event(type_=EventType.ERRATUM, id=erratum_id)
 
-                ctx.save_erratum_job('event-', erratum_job)
+        # fetch erratum details (TODO: releases for now)
+        errata = Erratum.from_errata_tool(event)
 
-    else:
-        for erratum in ctx.load_initial_errata('init-'):
-            # fetch erratum details, namely releases
-            releases = ['RHEL-8.10.0', 'RHEL-9.4.0']
+        for erratum in errata:
+            erratum_job = ErratumJob(event=event, erratum=erratum)
 
-            for release in releases:
-                erratum_job = ErratumJob(event=erratum.event, erratum=Erratum(release=release))
-
-                ctx.save_erratum_job('event-', erratum_job)
+            ctx.save_erratum_job('event-', erratum_job)
 
 
 @main.command(name='jira')
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,6 +24,10 @@ repos:
           - "attrs>=20.3.0"
           - "ruamel.yaml>=0.16.6"
           - "jinja2>=2.11.3"
+          # prevent conflict between types-requests and urllib3
+          - "types-requests<2.31.0.7; python_version < '3.10'"
+          - "types-requests; python_version >= '3.10'"
+
         pass_filenames: false
         args: [--config-file=pyproject.toml]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,6 +75,9 @@ dependencies = [
     "mypy",
     "pytest",
     "pre-commit",
+    # prevent conflict between types-requests and urllib3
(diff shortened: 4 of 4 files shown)


The-Mule:
Here's an example how it works now:

```
❯ REQUESTS_CA_BUNDLE=/etc/pki/tls/cert.pem newa event -e 128049
[2024-03-04 15:59:10,919] [event   ] Erratum job 128049 @ RHEL-9.0.0.Z.EUS written to /home/omoris/Work/Projects/newa-omoris/state/event-128049-RHEL-9.0.0.Z.EUS.yaml
[2024-03-04 15:59:10,920] [event   ] Erratum job 128049 @ RHEL-9.2.0.Z.EUS written to /home/omoris/Work/Projects/newa-omoris/state/event-128049-RHEL-9.2.0.Z.EUS.yaml
[2024-03-04 15:59:10,920] [event   ] Erratum job 128049 @ RHEL-9.3.0.Z.MAIN written to /home/omoris/Work/Projects/newa-omoris/state/event-128049-RHEL-9.3.0.Z.MAIN.yaml
[2024-03-04 15:59:10,920] [event   ] Erratum job 128049 @ RHEL-8.6.0.Z.EUS written to /home/omoris/Work/Projects/newa-omoris/state/event-128049-RHEL-8.6.0.Z.EUS.yaml
[2024-03-04 15:59:10,921] [event   ] Erratum job 128049 @ RHEL-8.8.0.Z.EUS written to /home/omoris/Work/Projects/newa-omoris/state/event-128049-RHEL-8.8.0.Z.EUS.yaml
[2024-03-04 15:59:10,921] [event   ] Erratum job 128049 @ RHEL-8.9.0.Z.MAIN written to /home/omoris/Work/Projects/newa-omoris/state/event-128049-RHEL-8.9.0.Z.MAIN.yaml
[2024-03-04 15:59:10,921] [event   ] Erratum job 128049 @ RHEL-7.9.Z written to /home/omoris/Work/Projects/newa-omoris/state/event-128049-RHEL-7.9.Z.yaml
```

newa/__init__.py
```diff
@@ -1,14 +1,19 @@
...
 
 import attrs
 import jinja2
+import requests
 import ruamel.yaml
 import ruamel.yaml.nodes
 import ruamel.yaml.representer
+import time
 from attrs import define, field
+from requests_kerberos import HTTPKerberosAuth
```
happz, line 16: Both `requests` and `requests_kerberos` should be added to `pyproject.toml`, https://github.com/RedHatQE/newa/blob/main/pyproject.toml#L39

The-Mule: Ah, sorry - fixed now.

newa/__init__.py
```diff
@@ -129,6 +134,37 @@ class Event(Serializable):
...
+        attempts = 0
+        while attempts < max_attempts:
+            r = requests.get(url, auth=HTTPKerberosAuth(delegate=True))
+            if r.status_code == 200:
+                return r.json()
+            else:
+                time.sleep(delay)
+                attempts += 1
+
+        raise Exception(f"GET request to {url} failed")
```
happz, line 158: * does it have to be a static method of `ErrataTool`? It seems to accept a generic URL, which may or may not point to ET, the code seems to not care. A helper function similar to `render_template()` would serve us as well.
* I'd recommend renaming `max_attempts` to `attempts`, then you can `while attempts` and `attempts -= 1` at the end of each loop. One less variable to play with.
* the `else` keyword is not needed, because the `if` branch ends with `return` - if the condition was true, `if` happens and function ends. If it was false, then the code after `if` is executed, `else` is then not needed.

newa/__init__.py
```diff
@@ -145,13 +181,63 @@ class InitialErratum(Serializable):
...
+
+    # TODO: We might need to add more in the future.
     release: str
-    # builds: list[...] = ...
+    builds: list[str]
 
-    def fetch_details(self) -> None:
-        raise NotImplementedError
+    @classmethod
+    def from_errata_tool(cls, erratum_id : str) -> list[Erratum]:
```
happz, line 195: The input should probably be `Event`. It carries the ET ID, and it's the object that newa code uses to initialize errata structures from. So, when speaking about "erratum ID", the answer should be `event.id`.

newa/__init__.py
The-Mule: Yeah, I was on the edge on this and felt whole ErrataTool thing is something we don't really need and we can use the non-class methods instead. I'll move this one out and later we might drop whole ErrataTool thing if needed.

newa/__init__.py
The-Mule: Right. That sounds reasonable. I'll fix that.

kkaarreell:
While testing I have got
```
$ hatch -e dev shell
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [21 lines of output]
      /bin/sh: line 1: krb5-config: command not found
...
```
`krb5-devel` package was required, just thinking if we need to capture that dependency.

thrix:
@kkaarreell adding a patch for it ...

thrix:
@kkaarreell somethting like this we use for other projects:

https://github.com/RedHatQE/newa/pull/13

Just run `make` to see the nice help for the available target.

kkaarreell:
Would it be reasonable to define default cacert path in the program itself?

lukaszachy:
> Would it be reasonable to define default cacert path in the program itself?

IMO it could be part of hatch configuration https://hatch.pypa.io/1.9/config/environment/overview/#environment-variables

Once 'newa' uses system packages (namely python3-requests) it will use the system CA as downstream requests.certs is patched to point to the system CA instead of the 'certifi' package.

newa/cli.py
```diff
@@ -114,7 +114,7 @@ def cmd_event(ctx: CLIContext, errata_ids: tuple[str, ...]) -> None:
             releases = ['RHEL-8.10.0', 'RHEL-9.4.0']
```
kkaarreell, line 114: I think we should modify the `else` block here, hardcoding defaults doesn't seem appropriate now when we can fetch details from ET. In particular, in the `else` block we should read event details including `errata_id` from the `init-` file and then fetch details (unifying this part if the block above so we are not duplicating the code)

newa/cli.py
```diff
@@ -101,10 +101,10 @@ def cmd_event(ctx: CLIContext, errata_ids: tuple[str, ...]) -> None:
             event = Event(type_=EventType.ERRATUM, id=erratum_id)
```
kkaarreell, line 101: Does this `_` belong here? In the YAML it produces records like:
```
event:
  id: '128049'
  type_: erratum
```

newa/cli.py
kkaarreell: I would propose to use something like this

```
@click.pass_obj
def cmd_event(ctx: CLIContext, errata_ids: tuple[str, ...]) -> None:
    ctx.enter_command('event')

    # when errata_ids were provided on a cmdline
    # process errata_ids loaded from init- files
    if not errata_ids:
        errata_ids = []
        for erratum in ctx.load_initial_errata('init-'):
            errata_ids.append(erratum.event.id)

    for erratum_id in errata_ids:
        event = Event(type_=EventType.ERRATUM, id=erratum_id)

        # fetch erratum details, namely releases
        errata = Erratum.from_errata_tool(event)

        for erratum in errata:
            erratum_job = ErratumJob(event=event, erratum=erratum)

            ctx.save_erratum_job('event-', erratum_job)
```
We should probably add also some error handling when no ids are found at all.

newa/cli.py
happz: `type` is a Python built-in function, hence `type_` to avoid the conflict. And it's ugly and I'd be happy to rename the field to something better.

happz: I agree

newa/__init__.py
```diff
@@ -129,6 +147,24 @@ class Event(Serializable):
...
 
+@define
+class ErrataTool:
+    """
+    Interface to Errata Tool instance
+
+    Its only purpose is to serve information about an erratum and its builds.
+    """
+
+    url : str = "https://errata.devel.redhat.com"
```
happz, line 158: Is it a good idea to 1. hardcode the URL, 2. use an internal hostname in public code?

newa/cli.py
```diff
@@ -114,7 +114,7 @@ def cmd_event(ctx: CLIContext, errata_ids: tuple[str, ...]) -> None:
             releases = ['RHEL-8.10.0', 'RHEL-9.4.0']
 
             for release in releases:
-                erratum_job = ErratumJob(event=erratum.event, erratum=Erratum(release=release))
+                erratum_job = ErratumJob(event=erratum.event, erratum=Erratum(release=release, builds=[]))
```
happz, line 117: If you define `builds` with a default value, `builds: list[str] = dataclasses.field(default_factory=list)`, you may omit `builds=[]` here, an empty list will be constructed automagically.

newa/cli.py
The-Mule: I'll fix that.

newa/__init__.py
The-Mule: Good point, I assume there will be some kind of newa config at some point. For now, maybe we can souce url value from env variable?

newa/cli.py
The-Mule: Right, I think I tried that at some point but did it wrong way. I'll fix that.

newa/__init__.py
happz: Yep, that sounds like a good way till newa gets a config file.

newa/cli.py
The-Mule: Fixed.

newa/__init__.py
```diff
@@ -129,6 +148,26 @@ class Event(Serializable):
...
+class ErrataTool:
+    """
+    Interface to Errata Tool instance
+
+    Its only purpose is to serve information about an erratum and its builds.
+    """
+
+    # TODO: Until we have a dedicated newa config, we'll setup ET instance
+    # url via environment variable ET_URL
+    url: str = field(default=os.environ.get("ET_URL"))
```
kkaarreell, line 161: Despite it is a temporary solution, we can expect users would forget to do such a setup. I would propose to fail with a reasonable errro when ET_URL is not set.

happz:
@The-Mule I'd recommend installing `pre-commit` in your local repo, https://pre-commit.com/#3-install-the-git-hook-scripts, then run it. It's what is failing in the Github action, https://github.com/RedHatQE/newa/actions/runs/8340484916/job/22824507975?pr=12#step:4:68

newa/__init__.py
The-Mule: I added field validator that would expect ET_URL to match a specific regexp:

```
    url: str = field(default=os.environ.get("ET_URL", default=""),
                     validator=validators.matches_re("^https?://.+$"))
```

If ET_URL is missing or seems to be invalid `ValueError` exception will pop up:

```
❯ REQUESTS_CA_BUNDLE=/etc/pki/tls/cert.pem newa  event -e 128049
Traceback (most recent call last):
  File "/home/omoris/.local/share/hatch/env/virtual/newa/6kDqWrPU/dev/bin/newa", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/omoris/.local/share/hatch/env/virtual/newa/6kDqWrPU/dev/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omoris/.local/share/hatch/env/virtual/newa/6kDqWrPU/dev/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/omoris/.local/share/hatch/env/virtual/newa/6kDqWrPU/dev/lib/python3.11/site-packages/click/core.py", line 1719, in invoke
    rv.append(sub_ctx.command.invoke(sub_ctx))
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omoris/.local/share/hatch/env/virtual/newa/6kDqWrPU/dev/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omoris/.local/share/hatch/env/virtual/newa/6kDqWrPU/dev/lib/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omoris/.local/share/hatch/env/virtual/newa/6kDqWrPU/dev/lib/python3.11/site-packages/click/decorators.py", line 45, in new_func
    return f(get_current_context().obj, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omoris/Work/Projects/newa-omoris/newa/cli.py", line 111, in cmd_event
    errata = Erratum.from_errata_tool(event)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/omoris/Work/Projects/newa-omoris/newa/__init__.py", line 233, in from_errata_tool
    releases_json = ErrataTool().fetch_releases(event.id)
                    ^^^^^^^^^^^^
  File "<attrs generated init newa.ErrataTool>", line 5, in __init__
  File "/home/omoris/.local/share/hatch/env/virtual/newa/6kDqWrPU/dev/lib/python3.11/site-packages/attr/validators.py", line 146, in __call__
    raise ValueError(
ValueError: ("'url' must match regex '^https?://.+$' ('' doesn't)", Attribute(name='url', default='', validator=<matches_re validator for pattern re.compile('^https?://.+$')>, repr=True, eq=True, eq_key=None, order=True, order_key=None, hash=None, init=True, metadata=mappingproxy({}), type='str', converter=None, kw_only=False, inherited=False, on_setattr=None, alias='url'), re.compile('^https?://.+$'), '')
```

Will that do the trick? Or do you prefer something more specific (ie. an exception that directly mentions ET_URL)?

The-Mule:
@happz I resolved what I could in `pre-commit` report but mypy and ruff issues seems to be beyond my control.

happz:
> @happz I resolved what I could in `pre-commit` report but mypy and ruff issues seems to be beyond my control.

The following should help:

```
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
index ae8841f..02db314 100644
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,6 +24,10 @@ repos:
           - "attrs>=20.3.0"
           - "ruamel.yaml>=0.16.6"
           - "jinja2>=2.11.3"
+          # prevent conflict between types-requests and urllib3
+          - "types-requests<2.31.0.7; python_version < '3.10'"
+          - "types-requests; python_version >= '3.10'"
+
         pass_filenames: false
         args: [--config-file=pyproject.toml]
 
diff --git a/pyproject.toml b/pyproject.toml
index 2777316..8e194bb 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,6 +75,9 @@ dependencies = [
     "mypy",
     "pytest",
     "pre-commit",
+    # prevent conflict between types-requests and urllib3
+    "types-requests<2.31.0.7; python_version < '3.10'",
+    "types-requests; python_version >= '3.10'",
     ]
 
 [tool.hatch.envs.dev.scripts]
@@ -117,7 +120,9 @@ python_version = "3.9"
 files = ["newa/"]
 
 [[tool.mypy.overrides]]
-module = []
+module = [
+    "requests_kerberos.*"
+]
 ignore_missing_imports = true
 
 [tool.autopep8]
@@ -163,6 +168,7 @@ select = [
     "RUF",  # ruff
     ]
 ignore = [
+    "UP007",  # Use `X | Y` for type annotations
 ]
 
 [tool.ruff.isort]
```

newa/__init__.py
happz: I'd say the exception is correct: when there's no value provided, the default is used, and after that, the validator is applied, and `""` does not match the regular exception.

I'd suggest the following:

```python
     # TODO: Until we have a dedicated newa config, we'll setup ET instance
     # url via environment variable ET_URL
     def _url_factory() -> str:
         if 'ET_URL' in os.environ:
             return os.environ['ET_URL']

         raise Exception("ET_URL envvar is required.")

     url: str = field(factory=_url_factory,
                      validator=validators.matches_re("^https?://.+$"))
```

This way, we can tell user the `ET_URL` is required, validator will take care of validation, and we avoid the `""` which would be unusable anyway.

One more note: I'd recommend adding `NEWA_` prefix to the envvar name. Eventually, newa will be integrated with other tools, and it's very easy to add envvars for all those tools, it's also extremely hard to resolve any conflict should one of the tools claim `ET_URL` for its own goals, with a different meaning. By namespacing it, we can avoid any conflict with `ET_URL` required by TF or tmt (although they would consume `TESTING_FARM_ET_URL` and `TMT_ET_URL`, respectively, but you get the point, right? :)

The-Mule: Agreed. Code updated (`ET_URL` -> `NEWA_ET_URL`) and added default as suggested (I used `@url.default` decorator instead of factory, it should make no difference).

The-Mule:
@thrix Hopefully all issues were addressed now and github actions are finally green too (kudos to @happz for helping with that).

kkaarreell (approved):
Thank you!

happz (approved):

PR #15, opened 2024-03-15
Add minimal JiraJob code

kkaarreell:
This PR adds minimal code for JiraJob class and associated classes Issue and Recipe.

### DIFF
--- a/newa/__init__.py
+++ b/newa/__init__.py
@@ -148,7 +148,27 @@ class Erratum(Cloneable, Serializable):
     """ An eratum """
 
     release: str
-    # builds: list[...] = ...
+    builds: list[str] = []
+
+    def fetch_details(self) -> None:
+        raise NotImplementedError
+
+
+@define
+class Issue(Cloneable, Serializable):
+    """ A Jira issue """
+
+    issue: str
+
+    def fetch_details(self) -> None:
+        raise NotImplementedError
+
+
+@define
+class Recipe(Cloneable, Serializable):
+    """ A job recipe """
+
+    url: str
 
     def fetch_details(self) -> None:
         raise NotImplementedError
@@ -162,7 +182,7 @@ class Job(Cloneable, Serializable):
         converter=lambda x: x if isinstance(x, Event) else Event(**x),
         )
 
-    # issue: ...
+    # jira: ...
     # recipe: ...
     # test_job: ...
     # job_result: ...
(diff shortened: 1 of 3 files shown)


kkaarreell:
`$ REQUESTS_CA_BUNDLE=/etc/pki/tls/cert.pem newa event jira` 
produces few `jira-` files like this one
```
$ cat jira-128049-RHEL-8.10.0-NEWA-12.yaml
erratum:
  builds: []
  release: RHEL-8.10.0
event:
  id: '128049'
  type_: erratum
jira:
  issue: NEWA-12
recipe:
  url: https://path/to/recipe.yaml
```
The new keys here are `jira` and `recipe`. This file is currently created only when `job_recipe` is defined in the configuration.

newa/cli.py
```diff
@@ -124,6 +142,15 @@ def cmd_event(ctx: CLIContext, errata_ids: tuple[str, ...]) -> None:
...
     ctx.enter_command('jira')
 
+    # This is just to provide fake Jira IDs until we can obtain real ones
+    def jira_issue_identifier() -> str:
+        num = 1
+        while True:
+            yield f'NEWA-{num}'
+            num += 1
+
+    jira_id_gen = jira_issue_identifier()
```
happz, line 152: https://docs.python.org/3/library/itertools.html#itertools.count comes to rescue :)

```python
jira_id_gen = itertools.count(start=1)
```

newa/__init__.py
```diff
@@ -148,7 +148,27 @@ class Erratum(Cloneable, Serializable):
     """ An eratum """
 
     release: str
-    # builds: list[...] = ...
+    builds: list[str] = []
```
happz, line 151: Just commented in another PR, but in general, `[]` will not work as expected, you need `factory` to spawn new list for every instance of this class: `builds: ... = field(factory=list)`.

newa/__init__.py
```diff
@@ -148,7 +148,27 @@ class Erratum(Cloneable, Serializable):
...
+
+    def fetch_details(self) -> None:
+        raise NotImplementedError
+
+
+@define
+class Issue(Cloneable, Serializable):
+    """ A Jira issue """
+
+    issue: str
```
happz, line 161: Maybe `id`? `issue.issue` looks like a typo, or a parent/child relationship, `issue.id` might be clearer.

kkaarreell: Right you are. I got confused by a conflict with `id` property but that one is defined for a different class, not this one.

newa/cli.py
kkaarreell: Updated

newa/__init__.py
kkaarreell: Hopefully fixed :-)

newa/__init__.py
kkaarreell: Changed.

happz (approved):

PR #20, opened 2024-03-25
Implementation of Jira command

### LINKED ISSUE #14: Ability to uniquely identify current and obsoleted Jira issue matching the respective item from Jira configuration
Let's suppose we have the following config item in the Jira processor configuration:
```
issues:
 - summary: "Testing ER#{errata.id} {errata.summary}"
   type: epic
   id: errata_epic
   on_respin: keep
-  summary: "Errata testing respin {errata.respin_count}"
   type: task
   id: errata_task
   parent: errata_epic
   on_respin: close
```
Now let's say we want to perform steps suggested by the configuration above for a respun advisory, in particular:
1. Search in Jira for epic to double-check if it already exists and create it if it does not
2. Search in Jira for the task so that it can be closed and create a new task.
The problem we are facing is: How to uniquely identify those new and obsoleted Jira issues so we can find them?

### LINKED ISSUE #17: Implement communication with Jira
At some point we need to start talking to Jira. We need to add functionality for that. We need to be able to search for issues, file issues and update them.

The-Mule:
Implement jira command and add all associated functionality.

Fixes #14 and #17.

### DIFF
--- a/newa/__init__.py
+++ b/newa/__init__.py
@@ -421,12 +421,12 @@ class Erratum(Cloneable, Serializable):
 
 @define
 class Issue(Cloneable, Serializable):
-    """ A Jira issue """
+    """ Issue - a key in Jira (eg. NEWA-123) """
 
-    id: str
+    id: str = field()
 
-    def fetch_details(self) -> None:
-        raise NotImplementedError
+    def __str__(self) -> str:
+        return self.id
 
 
 @define
--- a/newa/__init__.py
+++ b/newa/__init__.py
@@ -760,8 +760,9 @@ class ErratumConfig(Serializable):  # type: ignore[no-untyped-def]
 
 @frozen
 class IssueHandler:
-    """ An interface to Jira instance """
+    """ An interface to Jira instance handling a specific ErratumJob """
 
+    erratum_job: ErratumJob = field()
     url: str = field()
     token: str = field()
     project: str = field()
@@ -776,104 +777,146 @@ class IssueHandler:
         "field_epic_name": "customfield_12311141",
         }
 
-    _connection: jira.JIRA = field(init=False)
+    # NEWA identifiers related to ErratumJob
+    #
+    # 1. issues of all respins
+    partial_newa_id: str = field(init=False)
 
-    @_connection.default  # pyright: ignore [reportAttributeAccessIssue]
-    def _connection_factory(self) -> jira.JIRA:
+    # 2. issues of the current respin
+    newa_id: str = field(init=False)
+
+    # Actual Jira connection.
+    connection: jira.JIRA = field(init=False)
+
+    # NEWA label
+    newa_label: ClassVar[str] = "NEWA"
+
+    @partial_newa_id.default  # pyright: ignore [reportAttributeAccessIssue]
+    def partial_newa_id_factory(self) -> str:
+        return f"{IssueHandler.newa_label}: {self.erratum_job.id}"
+
+    @newa_id.default  # pyright: ignore [reportAttributeAccessIssue]
+    def _newa_id_factory(self) -> str:
+        return f"{IssueHandler.newa_label}: {self.erratum_job.id} " + \
+            f"({', '.join(sorted(self.erratum_job.erratum.builds))})"
+
+    @connection.default  # pyright: ignore [reportAttributeAccessIssue]
+    def connection_factory(self) -> jira.JIRA:
         return jira.JIRA(self.url, token_auth=self.token)
 
     def get_details(self, issue: Issue) -> jira.Issue:
         """ Return issue details """
 
         try:
-            return self._connection.issue(issue.id)
+            return self.connection.issue(issue.id)
         except jira.JIRAError as e:
             raise Exception(f"Jira issue {issue} not found!") from e
 
-    def search_open_issues(self, query: str) -> tuple[jira.client.ResultList[jira.Issue], str]:
-        """ Search for issues that are not closed (returns subset of fields, only) """
+    def get_open_issues(self,
+                        summary: str,
+                        all_respins: bool = False) -> jira.client.ResultList[jira.Issue]:
+        """
+        Get issues related to erratum job with given summary
+
+        Unless all_respins is defined only issues related to the last respin are returned.
+        """
 
         fields = ["summary", "description", "status", "resolution"]
 
+        newa_description = self.partial_newa_id if all_respins else self.newa_id
+
         query = \
-            f"({query}) AND " + \
-            f"project = '{self.project}' AND " + \
-            f"status not in ({','.join(self.transitions["closed"])})"
+            f"project = '{self.project}' AND labels in ({IssueHandler.newa_label}) AND " + \
+            f"summary ~ '{summary}' AND description ~ '{newa_description}' AND " + \
+            f"status not in ({','.join(self.transitions['closed'])})"
 
         # Result is always the right type with json_result=False (pyright just doesn't know that).
-        return self._connection.search_issues(query, fields=fields, json_result=False), query
+        return self.connection.search_issues(query, fields=fields, json_result=False)
 
-    def create_epic(self, data: dict[str, Any]) -> Issue:
-        """ Create Epic """
+    def create_issue(self,
+                     issue_type: IssueType,
+                     summary: str,
+                     description: str,
+                     assignee: str,
+                     parent: Issue | None) -> Issue:
+        """ Create issue """
 
-        data |= {
-            "issuetype": {"name": "Epic"},
-            IssueHandler.custom_field_map["field_epic_name"]: data["summary"],
+        data = {
+            "project": {"key": self.project},
+            "summary": summary,
+            "description": f"{self.newa_id}\n\n{description}",
+            "assignee": {"name": assignee},
             }
-        return self._create_issue(data)
 
-    def create_task(self, data: dict[str, Any], epic: Optional[Issue] = None) -> Issue:
-        """ Create Task """
+        if issue_type == IssueType.EPIC:
+            data |= {
+                "issuetype": {"name": "Epic"},
+                IssueHandler.custom_field_map["field_epic_name"]: data["summary"],
+                }
+        elif issue_type == IssueType.TASK:
+            data |= {"issuetype": {"name": "Task"}}
+            if parent:
+                data |= {IssueHandler.custom_field_map["field_epic_link"]: parent.id}
+        elif issue_type == IssueType.SUBTASK:
+            if not parent:
+                raise Exception("Missing task while creating sub-task!")
 
-        data |= {"issuetype": {"name": "Task"}}
-        if epic:
-            data |= {IssueHandler.custom_field_map["field_epic_link"]: epic.id}
-        return self._create_issue(data)
-
-    def create_subtask(self, data: dict[str, Any], task: Issue) -> Issue:
-        """ Create Subask """
-
-        data |= {
-            "issuetype": {"name": "Sub-task"},
-            "parent": {"key": task.id},
-            }
-        return self._create_issue(data)
-
-    def _create_issue(self, data: dict[str, Any]) -> Issue:
-        """ Create generic issue (label it with NEWA) """
-
-        data |= {"project": {"key": self.project}}
+            data |= {
+                "issuetype": {"name": "Sub-task"},
+                "parent": {"key": parent.id},
+                }
+        else:
+            raise Exception(f"Unknown issue type {issue_type}!")
 
         try:
-            jira_issue = self._connection.create_issue(data)
+            jira_issue = self.connection.create_issue(data)
             jira_issue.update(fields={"labels": [*jira_issue.fields.labels, 'NEWA']})
             return Issue(jira_issue.key)
         except jira.JIRAError as e:
             raise Exception("Unable to create issue!") from e
 
+    def refresh_issue(self, issue: Issue) -> None:
+        """ Update NEWA identifier of issue """
+
+        description = self.get_details(issue).fields.description
+
+        if isinstance(description, str) and self.partial_newa_id not in description:
+            raise Exception(f"Issue {issue} is missing NEWA identifier!")
+
+        if isinstance(description, str) and self.newa_id not in description:
+            new_description = re.sub(f"^{IssueHandler.newa_label}: .*\n",
+                                     f"{self.newa_id}\n", description)
+            try:
+                self.get_details(issue).update(fields={"description": new_description})
+                self.comment_issue(issue, "This issue was automatically refreshed by NEWA.")
+            except jira.JIRAError as e:
+                raise Exception(f"Unable to modify issue {issue}!") from e
+
     def comment_issue(self, issue: Issue, comment: str) -> None:
         """ Add comment to issue """
 
         try:
-            self._connection.add_comment(issue.id, comment)
+            self.connection.add_comment(issue.id, comment)
         except jira.JIRAError as e:
             raise Exception(f"Unable to add a comment to issue {issue}!") from e
-
-    def update_description(self, issue: Issue, description: str) -> None:
-        """ Update issue description """
-
-        try:
-            self.get_details(issue).update(fields={"description": description})
-        except jira.JIRAError as e:
-            raise Exception(f"Unable to modify issue {issue}!") from e
 
     def drop_obsoleted_issue(self, issue: Issue, obsoleted_by: Issue) -> None:
         """ Close obsoleted issue and link a new one """
 
         obsoleting_comment = \
             "This issue was automatically closed as " + \
-            f"{self.transitions["dropped"][0]} by NEWA " + \
-            f"(obsoleted by {obsoleted_by}."
+            f"{self.transitions['dropped'][0]} by NEWA " + \
+            f"(obsoleted by {obsoleted_by})."
 
         try:
-            self._connection.create_issue_link(type="relates to",
-                                               inwardIssue=issue.id,
-                                               outwardIssue=obsoleted_by.id,
-                                               comment={
+            self.connection.create_issue_link(type="relates to",
+                                              inwardIssue=issue.id,
+                                              outwardIssue=obsoleted_by.id,
+                                              comment={
                                                    "body": obsoleting_comment,
                                                    "visbility": None,
-                                                   })
-            self._connection.transition_issue(issue.id,
-                                              transition=self.transitions["dropped"][0])
+                                                  })
+            self.connection.transition_issue(issue.id,
+                                             transition=self.transitions["dropped"][0])
         except jira.JIRAError as e:
             raise Exception(f"Unable to close issue {issue}!") from e
--- a/newa/__init__.py
+++ b/newa/__init__.py
@@ -736,11 +736,11 @@ class OnRespinAction(Enum):
 class IssueAction:  # type: ignore[no-untyped-def]
     summary: str
     description: str
-    assignee: str
     id: str
     on_respin: Optional[OnRespinAction] = field(  # type: ignore[var-annotated]
         converter=lambda value: OnRespinAction(value) if value else None)
     type: IssueType = field(converter=IssueType)
+    assignee: Optional[str] = None
     parent_id: Optional[str] = None
     job_recipe: Optional[str] = None
     when: Optional[str] = None
@@ -809,14 +809,19 @@ class IssueHandler:
 
         Notice that Jira user name has various forms, it can be either an e-mail
         address or just an user name or even an user name with some sort of prefix.
+        It is possible that some e-mail addresses don't have Jira user associated,
+        e.g. some mailing lists. In that case empty string is returned.
         """
 
         if assignee_email not in self.user_names:
             assignee_names = [u.name for u in self.connection.search_users(user=assignee_email)]
-            if not assignee_names or len(assignee_names) > 1:
-                raise Exception(f"Exactly one Jira user is expected to match {assignee_email}"
+            if not assignee_names:
+                self.user_names[assignee_email] = ""
+            elif len(assignee_names) == 1:
+                self.user_names[assignee_email] = assignee_names[0]
+            else:
+                raise Exception(f"At most one Jira user is expected to match {assignee_email}"
                                 f"({', '.join(assignee_names)})!")
-            self.user_names[assignee_email] = assignee_names[0]
 
         return self.user_names[assignee_email]
 
@@ -876,16 +881,17 @@ class IssueHandler:
                      action: IssueAction,
                      summary: str,
                      description: str,
-                     assignee_email: str,
-                     parent: Issue | None) -> Issue:
+                     assignee_email: str | None = None,
+                     parent: Issue | None = None) -> Issue:
         """ Create issue """
 
         data = {
             "project": {"key": self.project},
             "summary": summary,
             "description": f"{self.newa_id(action)}\n\n{description}",
-            "assignee": {"name": self.get_user_name(assignee_email)},
             }
+        if assignee_email and self.get_user_name(assignee_email):
+            data |= {"assignee": {"name": self.get_user_name(assignee_email)}}
 
         if action.type == IssueType.EPIC:
             data |= {
--- a/newa/__init__.py
+++ b/newa/__init__.py
@@ -323,15 +323,10 @@ class Event(Serializable):
 
 @frozen
 class ErrataTool:
-    """
-    Interface to Errata Tool instance
-
-    Its only purpose is to serve information about an erratum and its builds.
-    """
+    """ Interface to Errata Tool instance """
 
     url: str = field(validator=validators.matches_re("^https?://.+$"))
 
-    # TODO: Not used at this point because we only consume builds now
     def fetch_info(self, erratum_id: str) -> JSON:
         return get_request(
             url=f"{self.url}/advisory/{erratum_id}.json",
@@ -358,9 +353,6 @@ class ErrataTool:
 
         errata = []
 
-        # TODO: We might need to assert QE (or later) state. There is no point
-        # in fetching errata in NEW_FILES where builds need not to be present.
-
         # In QE state there is are zero or more builds in an erratum, each
         # contains one or more packages, e.g.:
         # {
@@ -378,6 +370,7 @@ class ErrataTool:
         #   ]
         # }
 
+        info_json = self.fetch_info(event.id)
         releases_json = self.fetch_releases(event.id)
         for release in releases_json:
             builds = []
@@ -385,7 +378,14 @@ class ErrataTool:
             for item in builds_json:
                 builds += list(item.keys())
             if builds:
-                errata.append(Erratum(release, builds))
+                # ErrataTool tracks assignee by their e-mail but we only want the username.
+                assignee = info_json["people"]["assigned_to"].rsplit('@')[0]
+                errata.append(Erratum(id=event.id,
+                                      respin_count=int(info_json["respin_count"]),
+                                      summary=info_json["synopsis"],
+                                      people_assigned_to=assignee,
+                                      release=release,
+                                      builds=builds))
             else:
                 raise Exception(f"No builds found in ER#{event.id}")
 
@@ -414,9 +414,12 @@ class Erratum(Cloneable, Serializable):
     Represents a set of builds targetting a single release.
     """
 
-    # TODO: We might need to add more in the future.
-    release: str
-    builds: list[str] = field(factory=list)
+    id: str = field()
+    respin_count: int = field(repr=False)
+    summary: str = field(repr=False)
+    people_assigned_to: str = field(repr=False)
+    release: str = field()
+    builds: list[str] = field(factory=list, repr=lambda val: ','.join(sorted(val)))
 
 
 @define
--- a/newa/__init__.py
+++ b/newa/__init__.py
@@ -762,8 +762,12 @@ class ErratumConfig(Serializable):  # type: ignore[no-untyped-def]
 class IssueHandler:
     """ An interface to Jira instance """
 
-    connection: jira.JIRA = field(init=False)
+    url: str = field()
+    token: str = field()
     project: str = field()
+
+    # Each project can have different semantics of issue status.
+    transitions: dict[str, list[str]] = field()
 
     # We assume that all projects have the following two custom fields mapped
     # as follows.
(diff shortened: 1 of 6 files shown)


kkaarreell:
@The-Mule Hi, could you please update envvars names, adding `NEWA_` prefix?

kkaarreell (approved):

newa/__init__.py
```diff
@@ -394,3 +395,20 @@ class ErratumConfig(Serializable):
...
+    connection: JIRA
+
+    def __init__(self) -> None:
+        # TODO: In the future both both Jira url and token will be sourced
+        # from newa config.
+        try:
+            url = os.environ["NEWA_JIRA_URL"]
+            token = os.environ["NEWA_JIRA_TOKEN"]
+        except KeyError as e:
+            raise ValueError(f"{'and'.join(e.args)} not set!") from None
```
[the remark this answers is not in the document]
The-Mule: Fixed, I used the same approach as I did in `ErrataTool` and used attrs default factory:

```
@define
class JiraHandler:
    """ An interface to Jira instance """

    # TODO: Until we have a dedicated newa config, we'll setup JIRA connection
    # via environment variables NEWA_JIRA_URL and NEWA_JIRA_TOKEN
    connection: JIRA = field()

    @connection.default  # pyright: ignore [reportAttributeAccessIssue]
    def _connection_factory(self) -> JIRA:
        if "NEWA_JIRA_URL" not in os.environ:
            raise Exception("NEWA_JIRA_URL envvar is required.")
        if "NEWA_JIRA_TOKEN" not in os.environ:
            raise Exception("NEWA_JIRA_TOKEN envvar is required.")
        return JIRA(os.environ["NEWA_JIRA_URL"], token_auth=os.environ["NEWA_JIRA_TOKEN"])
```

The-Mule:
@kkaarreell @happz Folks, can you take a brief look? See also updated description (I need to test it tomorrow and fix what won't work but other than that should be more or less ready for review).

FYI: pre-commit failures are not related to this PR.

newa/__init__.py
```diff
@@ -31,9 +31,14 @@
 import ruamel.yaml.representer
 import urllib3.response
 from attrs import define, field, frozen, validators
+from jira import JIRA, JIRAError
+from jira import Issue as JiraIssue
```
kkaarreell, line 35: Just a thought but I find the amount of Jira/Issue related classes already confusing and having hard time understanding where it comes from. Would it make sense to import only `jira` module so that references like `jira.Issue` would be more "informative"?

newa/__init__.py
```diff
@@ -382,12 +399,31 @@ def from_errata_tool(cls, event: Event) -> list[Erratum]:
 
 @define
 class Issue(Cloneable, Serializable):
```
kkaarreell, line 401: I think I will keep an eye on this class because I am not sure we really need it. My understanding is that for any operation with Jira we would be using classes from the `jira` module and I am wondering if there would be cases where we end up only with `Issue` class without getting the `JiraIssue` class right away (e.g. in order to add a comment etc.). I am not saying we should drop it right away, just keep it in mind as an opportunity to reduce "confusion".

newa/cli.py
```diff
@@ -194,50 +193,99 @@ def cmd_event(ctx: CLIContext, errata_ids: list[str]) -> None:
 def cmd_jira(ctx: CLIContext, issue_config: str) -> None:
     ctx.enter_command('jira')
 
-    # this is here temporarily so we generate fake Jira issue IDs
-    jira_id_gen = itertools.count(start=1)
+    # Get control of Jira.
+    # TODO: Hardcoded project for now.
+    # It will be served from the component-config.yaml later.
+    jira = JiraHandler("BASEQESEC")
```
kkaarreell, line 199: I think it would help if we could get the config related PR in and adjust this one according to it. It would also make testing for users way more comfortable.

newa/cli.py
```diff
@@ -194,50 +193,99 @@ def cmd_event(ctx: CLIContext, errata_ids: list[str]) -> None:
...
             print()
 
-            if action.id in known_issues:
-                raise Exception(f'Issue "{action.id}" is already created!')
-
+            # Find existing issues in Jira related to erratum_job and action.id.
+            #
+            # We re looking for Jira issues such that:
+            #
+            # 1) summary matches rendered_summary (e.g. Errata respin 0) and
```
kkaarreell, line 228: Again, probably not something for now but in the future we should probably have functions helping building the search query for every EventType.

kkaarreell: Btw, let's remove that `'respin 0'` string as it is confusing. At first, we are searching for issues independent on a respin.

newa/cli.py
```diff
@@ -194,50 +193,99 @@ def cmd_event(ctx: CLIContext, errata_ids: list[str]) -> None:
...
+            query = f""" summary ~ '{rendered_summary}' and
+                         description ~ '{erratum_job.erratum!r}' and
+                         description ~ '{erratum_job.erratum.release!r}' """
+            for jira_issue in jira.search_open_issue(query):
+                desc = jira_issue.fields.description
+
+                # We found relevant issue in Jira, it is either related to the
+                # current respin or not.
+                #
+                # 1) If it is, it is not created again - it is skipped.
```
kkaarreell, line 240: I think we will need to modify issue processing a bit so that at first we spit issues into  categories "current" and "old" and then process each category individually. The reason for that is that when closing "old" issue we would most likely want to add there a comment referencing the "current" one.

newa/cli.py
```diff
@@ -194,50 +193,99 @@ def cmd_event(ctx: CLIContext, errata_ids: list[str]) -> None:
...
+                # 2) If it is not, then it is related to the previous respin(s), its fate is
+                # driven by action.on_respin - it is either closed or kept open.
+                else:
+                    if action.on_respin == "close":
+                        key = jira_issue.key
+                        jira.close_issue(key)
+                        print(f"Closed issue {key}")
+
+            # For debugging.
             if action.parent_id and action.parent_id not in known_issues:
```
kkaarreell, line 254: I think this is important and we should not be doing the above processing before this block.

kkaarreell (changes requested):

newa/__init__.py
The-Mule: Yup, that sounds like a good idea.

newa/__init__.py
The-Mule: It actually make sense to keep it IMO. The reason is that we often only need to work with Jira ID (key, eg. NEWA-123) and for that having an envelope object (Issue) is kinda handy. When needed the actual content can be loaded into into (as it is currently implemented).

newa/cli.py
The-Mule: I'll take a look into this.

newa/__init__.py
```diff
@@ -382,12 +399,31 @@ def from_errata_tool(cls, event: Event) -> list[Erratum]:
...
+    Issue is identified by its key in Jira (eg. NEWA-123).
+    We don't associate the details of the issue, i.e. content of Jira issue,
+    by default but only on demand because it requires data retrieval from
+    Jira and it is not always needed.
+    """
 
-    def fetch_details(self) -> None:
-        raise NotImplementedError
+    id: str = field()
+    _details: JiraIssue = field(repr=False,
```
happz, line 412: Must be `Optional[JiraIssue]` because the default is `None`.

The-Mule: Fixed.

newa/__init__.py
```diff
@@ -382,12 +399,31 @@ def from_errata_tool(cls, event: Event) -> list[Erratum]:
...
+    def details(self) -> JiraIssue | None:
+        return self._details if self._details else self._fetch_details()
+
+    def _fetch_details(self) -> JiraIssue | None:
+        try:
+            # TODO: Hardcoded project for now.
+            # It will be served from the component-config.yaml later.
+            return JiraHandler("BASEQESEC").get_issue(self.id)
+        except JIRAError:
+            return None
```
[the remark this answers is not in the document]
The-Mule: Right, I was considering an option that handle exists but is missing in Jira but that is indeed error state - added an exception instaed of return.

newa/cli.py
The-Mule: Agreed, I'll keep it in the current one and once we have config PR merged, I'll add jira project and project-specific fields/statuses mappings into the configuration.

newa/cli.py
The-Mule: Sounds reasonable but the split might complicate code a bit. I'll keep the original version commented out above the new one so that you can still keep eye on the original one. The thing is that I still need to go through the issue actions and 

- I need to remember ones for which I need to create issues  (in the subsequent loop)
- I need to find old issues and associate them with their issue action so that I know what to do with then (in the subsequent loop

So I will end up with three loops:

1. The original one making the split.
2. The one for creating issues.
3. The one for processing issue actions for old issues.

kkaarreell: I think se vám process issue config one by one. If it refers a parent se dont know, push it at the end of the queue, if not, we can proceed  handling the Jira content. Just realized, when pushing item at the end, we need to add some sort of Infinite loop detection.

newa/cli.py
The-Mule: This should be fixed in jinja template. I'll do that.

The-Mule: Fixed in the configuration sample, a bit ugly but should work.

newa/cli.py
The-Mule: Please check the current version of the command. Here's what I did:

1) _First, go through issue actions to see what issues from the previous respin(s) need to be processed._

The sole purpose of this loop is to find outdated issues and store them with their associated on_respin action.

2) _Second, iterate over issue actions, take one, if it's not possible to finish it, put it back at the end of the queue. Create and store issue otherwise._

This is basically the original loop - we create issues if we can or put action to the end and get back to it again later. endless loop is detected too.

3) _Third, iterate over issue actions for issues related to the previous respin(s) and act accordingly._

Handle the outdated issues detected in the first step. based on their on_respin action.

The-Mule:
FYI: The last version still not fully tested. Once we agree on overall structure you can expect some polishing based on testing.

newa/cli.py
```diff
@@ -194,63 +194,125 @@ def cmd_event(ctx: CLIContext, errata_ids: list[str]) -> None:
...
+                         description ~ '{erratum_job.erratum.release!r}' """
+            for jira_issue in jira.search_open_issue(query):
+                desc = jira_issue.fields.description
+
+                # We found relevant issue in Jira, it is either related to the
+                # current respin (new) or not (old).
+                #
+                # 1) If it is, we encountered error (either none or all issues must exist).
+                if isinstance(desc, str) and desc.find(repr(erratum_job.erratum.builds)):
+                    raise Exception(
```
kkaarreell, line 237: I believe it would be perfectly valid to run `newa` command multiple times for a single artifact and re-use already existing Jira issue. This could be e.g. because a user wants to re-run previously aborted tests.

kkaarreell:
I have a proposal for the issue processing loop. It is not "complete" but I think you get the point. It has the following advantages:
 1. there is just a single Jira search
 2. config isssue processing goes one by one so in cases it gets interrupted somehow you won't end up with some old issues being closed while new issues not being yet created.
Please check if you think it would work.
```
        # Iterate over issue actions. Take one, if it's not possible to finish it,
        # put it back at the end of the queue.
        issue_actions = config.issues[:]
        endless_loop_test = {}

        while issue_actions:
            action = issue_actions.pop(0)

            print(f'* Checking {action.type.name} issue:')
            print(f'     summary: {action.summary}')
            print()


            # first check if we know the parent already and if not, defer action processing for later
            if action.parent_id and action.parent_id not in known_issues:
                # check if the queue got shorter since we processed this action.id for the last time
                last_queue_length = endless_loop_test.get(action.id, 0)
                current_queue_length = len(issue_actions)
                if last_queue_length == current_queue_length:
                    raise Exception(f"Infinite loop detected for issue action.id = '{action.id}'")
                endless_loop_test[action.id] = current_queue_length
                print(f'     !! Parent issue, "{action.parent_id}", is unknown, will try later')
                print()

                issue_actions.append(action)
                continue

            print(f'* Processing {action.type.name} issue:')
            print(f'     summary: {action.summary}')
            print()

            known_issues[action.id] = True

            # Find existing issues in Jira related to erratum_job and action.id.
            #
            # We re looking for Jira issues such that:
            # 1) summary matches rendered_summary (e.g. Errata respin 0) and
            # 2) description contains erratum_job.erratum (e.g. Erratum(130704) and
            # 3) description contains erratum_job.erratum.release (e.g. Release(RHEL-8.8.0.Z.EUS))
            # right now we do not have it but I think we should also incorporate action.type into search
            
            query = f""" summary ~ '{render_template(action.summary, ERRATUM=erratum_job)}' and
                         description ~ '{erratum_job.erratum!r}' and
                         description ~ '{erratum_job.erratum.release!r}' """

            old_issues = []
            new_issues = []
            for jira_issue in jira.search_open_issue(query):

                desc = jira_issue.fields.description
                # We found relevant issue in Jira, it is either related to the
                # current respin (new) or not (old).

                if isinstance(desc, str) and desc.find(repr(erratum_job.erratum.builds)):
                    new_issues.extend(jira_issue)
                else:
                    old_issues.extend(jira_issue)

            # if when no new issue as been found, we need to create it
            if not new_issues:
                # create new issue
                # create a fake Issue object for now
                new_issues.append(Issue(id=f'NEWA-{next(jira_id_gen)}'))
            elif len(new_issues > 1):
                # for some reason we have found more Jira issues than expected,
                # we may want to do something with them
                print(f'Multiple new issues found in Jira for {action.id}')

            # now we have the new Jira issue we are going to work with
            # old issues are stored in the old_issues list, usually there should be
            # one or none of them but we should be able to somehow handle a situation 
            # when there were more of them
            new_issue = new_issues[0]

            # now process the respective config item (schedule tests, close old issues etc.)
```

newa/cli.py
The-Mule: Alright, this requires change. Also, thanks for the suggestion for the loop below. I guess I misunderstood the previous remark about creating new issues before the old are closed - I was under impression that we want to create all new one before handling all old ones which is most likely not what you meant (based on the loop suggestion). I must admit I don't like the idea of complete split (what I did in the current version) very much. So I am glad I can get back from it.

The-Mule: The last version should reflect your proposal with some rather cosmetic differences and some part of the code from the previous version. It might not be semantically 100% bullet-proof yet because I am just about to run it for the first time now. So please expect some additional fixes but I hope we have an agreement on the overall structure now :).

newa/cli.py
```diff
@@ -194,62 +193,150 @@ def cmd_event(ctx: CLIContext, errata_ids: list[str]) -> None:
...
+                if isinstance(desc, str) and desc.find(repr(erratum_job.erratum.builds)):
+                    new_issues.append(jira_issue.key)
+
+                # 2) If it is not, then it is related to the previous respin(s) and need
+                # to be stored for further processing later.
+                else:
+                    old_issues.append(jira_issue.key)
+
+            # We might need to re-use the old issue...
+            if old_issues and action.on_respin == OnRespinAction.KEEP:
```
kkaarreell, line 273: I am no sure this is correct. We should probably clarify the exactl semantics of `OnRespinAction.KEEP` and I would in fact propose to introduce `OnRespinAction.REUSE`, meaning:

- KEEP = do not close but leave it as it is, maybe just update it with a comment about a new issue
- REUSE = do not close but re-use it as a new issue

Either way, old issue(s) should be "reused" only when no new issue have been found. If we already have new issue for that particular jira config item, then we should simply use it.

The-Mule:
Switching back to draft since I did a a few bigger changes based on the testing testing results with jira stage.

newa/cli.py
```diff
@@ -194,62 +194,165 @@ def cmd_event(ctx: CLIContext, errata_ids: list[str]) -> None:
...
+            if action.parent_id and action.parent_id not in processed_actions:
+                queue_length = len(issue_actions)
+                last_queue_length = endless_loop_check.get(action.id, 0)
+                if last_queue_length == queue_length:
+                    raise Exception(f"Parent {action.parent_id} for {action.id} not found!")
 
-            if action.parent_id and action.parent_id not in known_issues:
-                print(f'     !! Parent issue, "{action.parent_id}", is unknown, will try later')
-                print()
+                endless_loop_check[action.id] = queue_length
```
kkaarreell, line 237: I believe we do not have `action.id` set when it is not present in a config file and it is not a mandatory key (I think it should not be). Therefore, we should probably generate some uniqu `id` in the class constructor if it is not provided as an argument.

The-Mule:
@kkaarreell 

Alright, this should be hopefully very close to the final version if not final. I did a fair bit of testing with Jira stage and everything should be in order. I rebased all of it and moved related changes together so it might be better to review it again from the beginning. Here's the summary (matching 6 commits in reverse order):

1. Issue class was in the end very hand, we always need to store just an issue key and nothing else. I dropped fetching details because it is never needed (and will not be) in Issue, if one need to get to the issue details, it can be done by `IssueHandled.get_details(issue: Issue) -> jira.Issue`.
2. Erratum class was extended to hold other useful attributes that are finally needed in jira command - id, respin_count, summary, description, qe_assignee (previously it held only builds and release).
3. Component configuration was extended to capture other setting than issues - project (name) and transitions (lists of lists). Transitions are essential to be able to search and drop issues because the meaning of "closed" and "dropped" issue differs by projects. We always need to understand what "dropped" means (to what single transition it is mapped) and what "closed" means (to what transitions it is mapped) in a project. For instance, in BASEQESEC project  "closed" means one of "Done, Dropped or Complete" while in CRYPTO project "closed" means just "Closed".  In BASEQESEC "dropped" maps to "Dropped" but in CRYPTO it maps to "Won't Do". Configuration sample was updated too. 
4. Jira dependency added to the project (this is not new).
5. JiraHandler is now called IssueHandler to better match naming in the project (it handles Issues). Most notable changes are as follows. 
  a) custom fields is now class attribute and it is fixed, we assume that fields there are available in every project that we will use). 
  b) transitions is attribute defining transitions (see above). 
  c) search_open_issues now better handle issues that are not yet "closed" (where transitions define the meaning of it) and return the search query for inspection and debugging purposes. 
  d) close_issue was renamed to drop_issue because it will be only used for "dropping" (again, transitions define the meaning of it). We can add other closing functionality once it is needed.
6. A few logic bugs found during the testing were fixed amd prints were replace by logger calls but overall structure match the previous version.

I did the testing simulating normal workflow (respin 1 - initial version), respins (epic is re-used, other tickets are dropped and re-created) and some corners case (e.g. no respin happened but some issue was missing so that it is re-created).

The-Mule:
Here's the brief testing showcase where you can see what is exactly happening:

[testing.txt](https://github.com/RedHatQE/newa/files/15146598/testing.txt)

newa/cli.py
The-Mule: Alright, I'll come up with some default in IssueAction class (although for better error handling it is rather useful to have it).

newa/cli.py
The-Mule: This should be already handled I believe, when there are some KEEP old issues, they are added to new issues list (and removed from old issues list). So when there is already something in new issues list we will end up with multiple issues in that list and that raises an Exception (because of new issues we only allow nothing there or exactly one issue).

newa/cli.py
The-Mule: Fixed in the current PR.

newa/cli.py
The-Mule: All fixed in the current version.

newa/cli.py
The-Mule: Fixed in the current version.

The-Mule:
Based on offline feedback from @kkaarreell I added a couple more updates (arranged most of them to the original 6 commits and added one more). Here is the list of changes related to the commits (in reverse order):

1. no change
2. no change
3. no change
4. no change
5. added methods  `update_description`, `comment_issue` and change `drop_issue` to `drop_obsoleted_issue`. Former two are used in (6) and the last one is change to add comment and link to the new issues while dropping the old one. Also NEWA label is now added to all created issues.
6. Changes around NEWA identifier (requires (7)), issues are now identified by `NEWA: erratum_job.id` and when an issue is re-used old identifier is replaced by a new one and comment is added.
7. Added sortted builds of erratum into ErratumJob.id and added ErratumJob.partial_id property to be able to capture the original part without the builds (it is useful to detect the old issues).

The-Mule:
Now when config PR is merged, I am going to rebase onto it and ditch jira-related env-vars.

The-Mule:
Rebased onto current main (config support) and added another commit:

8. Use config settings instead of envvars in IssueHandler

newa/cli.py
kkaarreell: ack

newa/__init__.py
```diff
@@ -719,7 +730,131 @@ class IssueAction:  # type: ignore[no-untyped-def]
...
+
+    def search_open_issues(self, query: str) -> tuple[jira.client.ResultList[jira.Issue], str]:
+        """ Search for issues that are not closed (returns subset of fields, only) """
+
+        fields = ["summary", "description", "status", "resolution"]
+
+        query = \
+            f"({query}) AND " + \
+            f"project = '{self.project}' AND " + \
+            f"status not in ({','.join(self.transitions["closed"])})"
```
kkaarreell, line 782: When new issues have `newa` label, we should include it in the query, right?

newa/cli.py
```diff
@@ -205,62 +206,193 @@ def cmd_event(ctx: CLIContext, errata_ids: list[str]) -> None:
...
 
     for erratum_job in ctx.load_erratum_jobs('event-'):
+
         # read Jira issue configuration
         config = ErratumConfig.from_yaml_file(Path(os.path.expandvars(issue_config)))
 
-        # TODO: record created/existing issues. Instead of `Any`, maybe something
-        # from the Jira library would be stored. Or just a Jira ticket ID.
-        known_issues: dict[str, Any] = {}
+        jira = IssueHandler(jira_url, jira_token, config.project, config.transitions)
```
kkaarreell, line 222: Followup on a conversation from issue#22, do you think it makes sense to read project settings from a config file, envvar, jira subcommand `--project` option etc? I did something similar in
https://github.com/RedHatQE/newa/pull/27/files#diff-fbfe615aa583117b7a57302f5da6ff45c1eda26a6f4b2406b68a09ffe1bd698aR383
but TBH I am not sure which settings should have priority. Or would you rather "enforce" this kind of settings for the issue config file?

newa/__init__.py
The-Mule: Right, right, to be honest I am actually even thinking about moving all the search logic (including the stuff around identifier into the IssueHandler itself and creating an association between IssueHandler and ErratumJob. We only need to work with newa issue identifiers when working with Jira and hence this approach would reduce the code in jira command and make it a bit more readable. It would be a separate commit so that we can pick either version. That is most likely the last thing I have on my mind (it can be also done separately).

newa/cli.py
The-Mule: I considered it before making this change and here's the situation a bit more complicate because of `transitions` - it would be hard (or at least too verbose) to express resolution transitions (that is how jira calls it) on a command line or via envvar. So that part would end up in component config anyway and hence moving project out of it won't help too much. 

I'd propose to keep it all there (in a component config) for now and later we try to see what resolution transitions we will be working. If there is some reasonable default for most of the projects then we can assume that default and then we would only need `transitions` setting for special cases. In that case having `project` configurable either by an option of by envvar would make more sense.

newa/__init__.py
The-Mule: Now IssueHandler is very generic and I don't thing that is needed. We only use it to map IssueActions to jira and work with them - all the actions are using newa identifier and so on. So it kinda make more sense to move the identifier logic into it and provide methods for search all issues related to specific errata job (all respins or current one or the old ones), we current we this code in jira command.

kkaarreell: Right, I think I rather meant adding it to the query which is later passed here as an argument.

newa/cli.py
kkaarreell: Ack

newa/__init__.py
The-Mule: Sure, I'll add that. Previous comments were just a heads-up that it will be a part of bigger update. I'll get to that later today and that should conclude this PR.

The-Mule: Done, I added one commit to move all Jira logic to IssueHandler, it made jira command implementation way more readable.

The-Mule:
Based on the last (review from @kkaarreell  (via private channel) I am adding [00bd11a](https://github.com/RedHatQE/newa/pull/20/commits/00bd11a61faf3bfefc3d34d8052b69218c653016). Please see commit message for more details.

kkaarreell:
> Based on the last (review from @kkaarreell (via private channel) I am adding [00bd11a](https://github.com/RedHatQE/newa/pull/20/commits/00bd11a61faf3bfefc3d34d8052b69218c653016). Please see commit message for more details.

You will hate me for this but what about searching for Jira user using the email?
https://jira.readthedocs.io/api.html#jira.client.JIRA.search_users

The-Mule:
Not all all! Thanks for the hint, I did a few tries and this method does exactly what we need. We can also get rid of a hack I had to make in ErrataTool class before. Let me rebase the last commit.

The-Mule:
@kkaarreell I updated the last commit to cache Jira user names so that they can be re-used without querying the connection again (also, I moved a comment explaining newa_id() into docstring).

newa/cli.py
```diff
@@ -206,75 +206,153 @@ def cmd_event(ctx: CLIContext, errata_ids: list[str]) -> None:
...
+                #
+                # However, it might happen that we encounter subtask issue that is new but its
+                # original parent task got dropped (by human mistake, newa would never do that).
+                # By this time new parent task already exists. Unfortunately, Jira REST API does
+                # not allow updating 'parent' field [1] and hence we cannot re-use the issue with
+                # updated parent - we need to handle it as an old one (unless it has KEEP on_respin
+                # action it will get dropped and new one is created with the proper parent).
+                #
+                # [1] https://jira.atlassian.com/browse/JRASERVER-68763
+                decision_result = ""
```
happz, line 288: This variable looks more like a `is_new`, a boolean flag. Is it necessary to express the new/old state with strings? May there be more states/decisions?

newa/cli.py
```diff
@@ -206,75 +206,153 @@ def cmd_event(ctx: CLIContext, errata_ids: list[str]) -> None:
...
+
+            # Old issue(s) can be re-used for the current respin.
+            if old_issues and action.on_respin == OnRespinAction.KEEP:
+                new_issues.extend(old_issues)
+                old_issues = []
+
+            # Processing new issues.
+            #
+            # 1. Either there is no new issue (it does not exist yet - we need to create it).
+            if len(new_issues) == 0:
```
happz, line 312: Python idiom for testing emptiness of a list: `if not new_issues`

newa/__init__.py
```diff
@@ -742,7 +747,219 @@ class IssueAction:  # type: ignore[no-untyped-def]
...
+            result[jira_issue["key"]] = {"description": jira_issue["fields"]["description"]}
+            if "parent" in jira_issue["fields"]:
+                result[jira_issue["key"]] |= {"parent": jira_issue["fields"]["parent"]["key"]}
+        return result
+
+    def create_issue(self,
+                     action: IssueAction,
+                     summary: str,
+                     description: str,
+                     assignee_email: str,
```
kkaarreell, line 880: Can  we make assignee optional? In Jira it is quite common to have issues unassigned, however current implementation requires it.
```
Traceback (most recent call last):
  File "/home/ksrot/.local/share/hatch/env/virtual/newa/j_Eh7vfQ/dev/bin/newa", line 8, in <module>
    sys.exit(main())
  File "/home/ksrot/.local/share/hatch/env/virtual/newa/j_Eh7vfQ/dev/lib/python3.9/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/home/ksrot/.local/share/hatch/env/virtual/newa/j_Eh7vfQ/dev/lib/python3.9/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/home/ksrot/.local/share/hatch/env/virtual/newa/j_Eh7vfQ/dev/lib/python3.9/site-packages/click/core.py", line 1719, in invoke
    rv.append(sub_ctx.command.invoke(sub_ctx))
  File "/home/ksrot/.local/share/hatch/env/virtual/newa/j_Eh7vfQ/dev/lib/python3.9/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/ksrot/.local/share/hatch/env/virtual/newa/j_Eh7vfQ/dev/lib/python3.9/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/home/ksrot/.local/share/hatch/env/virtual/newa/j_Eh7vfQ/dev/lib/python3.9/site-packages/click/decorators.py", line 45, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/home/ksrot/devel/github/newa/newa/cli.py", line 220, in cmd_jira
    config = ErratumConfig.from_yaml_file(Path(os.path.expandvars(issue_config)))
  File "/home/ksrot/devel/github/newa/newa/__init__.py", line 312, in from_yaml_file
    return cls.from_yaml(filepath.read_text())
  File "/home/ksrot/devel/github/newa/newa/__init__.py", line 308, in from_yaml
    return cls(**data)
  File "<attrs generated init newa.ErratumConfig>", line 6, in __init__
  File "/home/ksrot/devel/github/newa/newa/__init__.py", line 755, in <lambda>
    factory=list, converter=lambda issues: [
  File "/home/ksrot/devel/github/newa/newa/__init__.py", line 756, in <listcomp>
    IssueAction(**issue) for issue in issues])
TypeError: __init__() missing 1 required positional argument: 'assignee'
```

newa/__init__.py
```diff
@@ -742,7 +747,219 @@ class IssueAction:  # type: ignore[no-untyped-def]
...
+    def newa_id(self, action: IssueAction, partial: bool = False) -> str:
+        """
+        NEWA identifier
+
+        Construct so-called NEWA identifier - it identifies all issues of given
+        action for errata. By default it defines issues related to the current
+        respin. If 'partial' is defined it defines issues relevant for all respins.
+        """
+
+        newa_id = f"{IssueHandler.newa_label} {action.id}: {self.erratum_job.id}"
```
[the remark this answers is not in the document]
The-Mule: I would have to escape # (because Jira interprets it as ordered list). Can we do this?

`::: NEWA errata_epic: E: 130704 @ RHEL-8.8.0.Z.EUS (libreswan-4.9-3.el8_8.1) :::`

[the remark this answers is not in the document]
kkaarreell: Works for me.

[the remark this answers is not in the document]
The-Mule: Fixed.

newa/cli.py
The-Mule: Fixed.

newa/cli.py
The-Mule: Fixed (also - that condition block was made simpler).

newa/__init__.py
The-Mule: Fixed (and tested).

kkaarreell (approved):
Can't comment on the code quality but I am happy with this PR functionality-wise. Thank you.

happz (approved):

PR #21, opened 2024-03-25
Add minimal implementation of the 'schedule' subcommand.

kkaarreell:
This PR adds minimal code for 'schedule' subcommand, implementing associated classes RecipeConfig, Request and RequestJob and essential expansion of the RecipeConfig into individual combinations (Requests).

### DIFF
--- a/newa/__init__.py
+++ b/newa/__init__.py
@@ -1,12 +1,26 @@
 from __future__ import annotations
 
+import copy
 import io
+import itertools
 import os
 import re
 import time
+from collections.abc import Iterator
 from enum import Enum
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Optional, TypeVar, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Literal,
+    Optional,
+    TypedDict,
+    TypeVar,
+    Union,
+    cast,
+    overload,
+    )
 
 import attrs
 import jinja2
@@ -14,6 +28,7 @@ import requests
 import ruamel.yaml
 import ruamel.yaml.nodes
 import ruamel.yaml.representer
+import urllib3.response
 from attrs import define, field, frozen, validators
 from requests_kerberos import HTTPKerberosAuth
 
@@ -21,6 +36,7 @@ if TYPE_CHECKING:
     from typing_extensions import Self, TypeAlias
 
     ErratumId: TypeAlias = str
+    JSON: TypeAlias = Any
 
 
 T = TypeVar('T')
@@ -89,13 +105,71 @@ def render_template(
         raise Exception("Could not render template.") from exc
 
 
-def krb_get_request(url: str, attempts: int = 5, delay: int = 5) -> Any:
-    """ Generic GET request using Kerberos authentication """
+class ResponseContentType(Enum):
+    TEXT = 'text'
+    JSON = 'json'
+    RAW = 'raw'
+    BINARY = 'binary'
 
+
+@overload
+def get_request(
+        *,
+        url: str,
+        krb: bool = False,
+        attempts: int = 5,
+        delay: int = 5,
+        response_content: Literal[ResponseContentType.TEXT]) -> str:
+    pass
+
+
+@overload
+def get_request(
+        *,
+        url: str,
+        krb: bool = False,
+        attempts: int = 5,
+        delay: int = 5,
+        response_content: Literal[ResponseContentType.BINARY]) -> bytes:
+    pass
+
+
+@overload
+def get_request(
+        *,
+        url: str,
+        krb: bool = False,
+        attempts: int = 5,
+        delay: int = 5,
+        response_content: Literal[ResponseContentType.JSON]) -> JSON:
+    pass
+
+
+@overload
+def get_request(
+        *,
+        url: str,
+        krb: bool = False,
+        attempts: int = 5,
+        delay: int = 5,
+        response_content: Literal[ResponseContentType.RAW]) -> urllib3.response.HTTPResponse:
+    pass
+
+
+def get_request(
+        url: str,
+        krb: bool = False,
+        attempts: int = 5,
+        delay: int = 5,
+        response_content: ResponseContentType = ResponseContentType.TEXT) -> Any:
+    """ Generic GET request, optionally using Kerberos authentication """
     while attempts:
-        r = requests.get(url, auth=HTTPKerberosAuth(delegate=True))
+        r = requests.get(url, auth=HTTPKerberosAuth(delegate=True)) if krb else requests.get(url)
         if r.status_code == 200:
-            return r.json()
+            response = getattr(r, response_content.value)
+            if callable(response):
+                return response()
+            return response
         time.sleep(delay)
         attempts -= 1
 
@@ -184,6 +258,11 @@ class Serializable:
     def from_yaml_file(cls: type[SerializableT], filepath: Path) -> SerializableT:
         return cls.from_yaml(filepath.read_text())
 
+    @classmethod
+    def from_yaml_url(cls: type[SerializableT], url: str) -> SerializableT:
+        r = get_request(url=url, response_content=ResponseContentType.TEXT)
+        return cls.from_yaml(r)
+
 
 @define
 class Event(Serializable):
@@ -212,11 +291,17 @@ class ErrataTool:
         raise Exception("NEWA_ET_URL envvar is required.")
 
     # TODO: Not used at this point because we only consume builds now
-    def fetch_info(self, erratum_id: str) -> Any:
-        return krb_get_request(f"{self.url}/advisory/{erratum_id}.json")
+    def fetch_info(self, erratum_id: str) -> JSON:
+        return get_request(
+            url=f"{self.url}/advisory/{erratum_id}.json",
+            krb=True,
+            response_content=ResponseContentType.JSON)
 
-    def fetch_releases(self, erratum_id: str) -> Any:
-        return krb_get_request(f"{self.url}/advisory/{erratum_id}/builds.json")
+    def fetch_releases(self, erratum_id: str) -> JSON:
+        return get_request(
+            url=f"{self.url}/advisory/{erratum_id}/builds.json",
+            krb=True,
+            response_content=ResponseContentType.JSON)
 
 
 @define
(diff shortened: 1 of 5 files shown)


The-Mule:
@kkaarreell FYI, see [this](https://github.com/RedHatQE/newa/pull/12#issuecomment-2012185177) for for resolving the failed pre-commit actions check.

kkaarreell:
I have updated PR with simple expansion of a recipe. Example:

```
REQUESTS_CA_BUNDLE=/etc/pki/tls/cert.pem newa event -e 128049 jira schedule
```
not generates files like this one:
```
$ cat state/request-128049-RHEL-9.4.0-NEWA-6-REQ-4.yaml
erratum:
  builds: []
  release: RHEL-9.4.0
event:
  id: '128049'
  type_: erratum
jira:
  id: NEWA-6
recipe:
  url: https://raw.githubusercontent.com/RedHatQE/newa/ks_recipe_job/component-recipe.yaml.sample
request:
  context:
    arch: aarch64
    fips: no
    tier: 1
  environment:
    DESCRIPTION: This is a test recipe
    FIPS: FIPS NOT ENABLED
  git_ref: main
  git_url: https://github.com/RedHatQE/newa.git
  id: REQ-4
  tmt_path: ''
```

newa/__init__.py
```diff
@@ -311,6 +335,62 @@ class Recipe(Cloneable, Serializable):
...
+@define
+class Request(Cloneable, Serializable):
+    """ A test job request configuration """
+
+    id: str
+    context: Optional[dict] = {}
+    environment: Optional[dict] = {}
+    git_url: Optional[str] = ''
+    git_ref: Optional[str] = ''
+    tmt_path: Optional[str] = ''
```
happz, line 388: A couple of notes WRT default values:

* `Optional[dict] = {}` - would `Optional` ever apply? The attribute would start with a genuine dictionary and does not seem to be set to `None`, ever.
* mutable defaults are one of Python's traps, try to avoid them. Otherwise, all `Request` instances would share the very same dictionary... The proper way is to use `factory`, i.e. `context: dict = field(factory=dict)` - every time an instance is created, `factory` callable is called and produces a dictionary specific for that instance only.
* `dict` lacks types annotating key and value. According to examples, I'd say `dict[str, str]` is the right type for context and environment.
* `Optional[str] = ''` - same question, `Optional` does suggest the attribute might be `None` from time to time, but that does not seem to be the case.
* But, `Optional[str] = ''` - an empty string as an "unset" value is hard to follow for linters. For mypy, it's not possible to tell whether a string is an empty string or not, it's always `str` and therefore the value is always "set" - let's say we have a function that runs `git clone` for a given URL and requires a valid URL. mypy will not stop you from calling such a function with an empty string, *you* would have to remember to check for such a case. Therefore, `None` is the best choice for "unset" values, and `git_url: Optional[str] = None` would be my pick for the winner.

newa/__init__.py
```diff
@@ -89,13 +91,24 @@ def render_template(
         raise Exception("Could not render template.") from exc
 
 
-def krb_get_request(url: str, attempts: int = 5, delay: int = 5) -> Any:
+def get_request(
+        url: str,
+        krb: bool = False,
+        attempts: int = 5,
+        delay: int = 5,
+        response_content: str = 'text') -> Any:
```
happz, line 99: It's better to use enums, linters can prevent you from making typos. With plain strings, nothing will prevent `get_request(..., response_content='tex')`.

```python
class ResponseContentType(enum.Enum):
  TEXT = 'text'
  JSON = 'json'
  ...

def get_request(..., response_content: ResponseContentType = ResponseContentType.TEXT)
  ...

  response = getattr(r, response_content.value)
```

No need for `if response_content in content_options` or `raise Exception` - mypy will prevent you from ever using a `response_content` value that's not in the enum.

newa/__init__.py
```diff
@@ -311,6 +335,62 @@ class Recipe(Cloneable, Serializable):
     url: str
 
 
+@define
+class RecipeConfig(Cloneable, Serializable):
+    """ A job recipe configuration """
+
+    fixtures: dict = {}
+    dimensions: dict = {}
```
happz, line 343: * mutable defaults require `factory`, `field(factory=dict)`.
* types for keys and values are needed. Keys seem to be strings, values are what, lists of mappings? I think this deserves some `TypedDict` definitions, or even actual classes, but at least `TypedDict` to define "context" and "environment" bits so we could then re-use and compose in `dimensions` and other complex attributes. Right now, it's very open.

newa/__init__.py
kkaarreell: Thank you, that's exactly what I was looking for. Just pushed an update.

newa/__init__.py
kkaarreell: Yeah, I thought you will tell this. :-) Updated.

newa/__init__.py
kkaarreell: I have added more types, making pre-commit happy. I am afraid that further improvements will require your personal involvement. :-)

kkaarreell: @happz Hi, would you find the current state OK-ish? :-)

happz: Checking... Although I will probably come with a commit, because I think we should be very careful with `Any` :)

happz:
@kkaarreell would mind trying this PR out with the following extra patch?

```diff
diff --git a/newa/__init__.py b/newa/__init__.py
index 02e6462..b5c0355 100644
--- a/newa/__init__.py
+++ b/newa/__init__.py
@@ -6,9 +6,21 @@ import itertools
 import os
 import re
 import time
+from collections.abc import Iterator
 from enum import Enum
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Optional, TypeVar, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Literal,
+    Optional,
+    TypedDict,
+    TypeVar,
+    Union,
+    cast,
+    overload,
+    )
 
 import attrs
 import jinja2
@@ -16,6 +28,7 @@ import requests
 import ruamel.yaml
 import ruamel.yaml.nodes
 import ruamel.yaml.representer
+import urllib3.response
 from attrs import define, field, frozen, validators
 from requests_kerberos import HTTPKerberosAuth
 
@@ -23,6 +36,7 @@ if TYPE_CHECKING:
     from typing_extensions import Self, TypeAlias
 
     ErratumId: TypeAlias = str
+    JSON: TypeAlias = Any
 
 
 T = TypeVar('T')
@@ -98,6 +112,50 @@ class ResponseContentType(Enum):
     BINARY = 'binary'
 
 
+@overload
+def get_request(
+        *,
+        url: str,
+        krb: bool = False,
+        attempts: int = 5,
+        delay: int = 5,
+        response_content: Literal[ResponseContentType.TEXT]) -> str:
+    pass
+
+
+@overload
+def get_request(
+        *,
+        url: str,
+        krb: bool = False,
+        attempts: int = 5,
+        delay: int = 5,
+        response_content: Literal[ResponseContentType.BINARY]) -> bytes:
+    pass
+
+
+@overload
+def get_request(
+        *,
+        url: str,
+        krb: bool = False,
+        attempts: int = 5,
+        delay: int = 5,
+        response_content: Literal[ResponseContentType.JSON]) -> JSON:
+    pass
+
+
+@overload
+def get_request(
+        *,
+        url: str,
+        krb: bool = False,
+        attempts: int = 5,
+        delay: int = 5,
+        response_content: Literal[ResponseContentType.RAW]) -> urllib3.response.HTTPResponse:
+    pass
+
+
 def get_request(
         url: str,
         krb: bool = False,
@@ -202,7 +260,7 @@ class Serializable:
 
     @classmethod
     def from_yaml_url(cls: type[SerializableT], url: str) -> SerializableT:
-        r = get_request(url, response_content=ResponseContentType.TEXT)
+        r = get_request(url=url, response_content=ResponseContentType.TEXT)
         return cls.from_yaml(r)
 
 
@@ -233,15 +291,15 @@ class ErrataTool:
         raise Exception("NEWA_ET_URL envvar is required.")
 
     # TODO: Not used at this point because we only consume builds now
-    def fetch_info(self, erratum_id: str) -> Any:
+    def fetch_info(self, erratum_id: str) -> JSON:
         return get_request(
-            f"{self.url}/advisory/{erratum_id}.json",
+            url=f"{self.url}/advisory/{erratum_id}.json",
             krb=True,
             response_content=ResponseContentType.JSON)
 
-    def fetch_releases(self, erratum_id: str) -> Any:
+    def fetch_releases(self, erratum_id: str) -> JSON:
         return get_request(
-            f"{self.url}/advisory/{erratum_id}/builds.json",
+            url=f"{self.url}/advisory/{erratum_id}/builds.json",
             krb=True,
             response_content=ResponseContentType.JSON)
 
@@ -338,20 +396,42 @@ class Recipe(Cloneable, Serializable):
     url: str
 
 
+# A tmt context for a recipe, dimension -> value mapping.
+RecipeContext = dict[str, str]
+
+# An environment for e recipe, name -> value mapping.
+RecipeEnvironment = dict[str, str]
+
+
+class RawRecipeConfigDimension(TypedDict, total=False):
+    context: RecipeContext
+    environment: RecipeEnvironment
+    git_url: Optional[str]
+    git_ref: Optional[str]
+
+
+_RecipeConfigDimensionKey = Literal['context', 'environment', 'git_url', 'git_ref']
+
+
+# A list of recipe config dimensions, as stored in a recipe config file.
+RawRecipeConfigDimensions = dict[str, list[RawRecipeConfigDimension]]
+
+
 @define
 class RecipeConfig(Cloneable, Serializable):
     """ A job recipe configuration """
 
-    fixtures: dict[str, dict[str, Any]] = field(factory=dict)
-    dimensions: dict[str, dict[str, Any]] = field(factory=dict)
-
-    def build_requests(self) -> list[Request]:
+    fixtures: RawRecipeConfigDimension = field(
+        factory=cast(Callable[[], RawRecipeConfigDimension], dict))
+    dimensions: RawRecipeConfigDimensions = field(
+        factory=cast(Callable[[], RawRecipeConfigDimensions], dict))
 
+    def build_requests(self) -> Iterator[Request]:
         # this is here to generate unique recipe IDs
         recipe_id_gen = itertools.count(start=1)
 
         # get all options from dimentions
-        options = []
+        options: list[list[RawRecipeConfigDimension]] = []
         for dimension in self.dimensions:
             options.append(self.dimensions[dimension])
         # generate combinations
@@ -360,23 +440,34 @@ class RecipeConfig(Cloneable, Serializable):
         for i in range(len(combinations)):
             combinations[i] = (self.fixtures,) + (combinations[i])
 
-        def merge_combination_data(combination: list[dict[str, Any]]) -> dict[str, Any]:
-            merged = {}
+        # Note: moved into its own function to avoid being indented too much;
+        # mypy needs to be silenced because we use `key` variable instead of
+        # literal keys defined in the corresponding typeddicts. And being nested
+        # too much, autopep8 was reformatting and misplacing `type: ignore`.
+        def _merge_key(
+                dest: RawRecipeConfigDimension,
+                src: RawRecipeConfigDimension,
+                key: str) -> None:
+            if key not in dest:
+                # we need to do a deep copy so we won't corrupt the original data
+                dest[key] = copy.deepcopy(src[key])  # type: ignore[literal-required]
+            elif isinstance(dest[key], dict) and isinstance(src[key], dict):  # type: ignore[literal-required]
+                dest[key].update(src[key])  # type: ignore[literal-required]
+            else:
+                raise Exception(f"Don't know how to merge record type {key}")
+
+        def merge_combination_data(
+                combination: tuple[RawRecipeConfigDimension, ...]) -> RawRecipeConfigDimension:
+            merged: RawRecipeConfigDimension = {}
             for record in combination:
                 for key in record:
-                    if key not in merged:
-                        # we need to do a deep copy so we won't corrupt the original data
-                        merged[key] = copy.deepcopy(record[key])
-                    elif isinstance(merged[key], dict) and isinstance(record[key], dict):
-                        merged[key].update(record[key])
-                    else:
-                        raise Exception(f"Don't know how to merge record type {key}")
+                    _merge_key(merged, record, key)
             return merged
 
         # now for each combination merge data from individual dimensions
         merged_combinations = list(map(merge_combination_data, combinations))
-        return [Request(id=f'REQ-{next(recipe_id_gen)}', **combination)
-                for combination in merged_combinations]
+        for combination in merged_combinations:
+            yield Request(id=f'REQ-{next(recipe_id_gen)}', **combination)
 
 
 @define
@@ -384,8 +475,8 @@ class Request(Cloneable, Serializable):
     """ A test job request configuration """
 
     id: str
-    context: dict[str, str] = field(factory=dict)
-    environment: dict[str, str] = field(factory=dict)
+    context: RecipeContext = field(factory=dict)
+    environment: RecipeEnvironment = field(factory=dict)
     git_url: Optional[str] = None
     git_ref: Optional[str] = None
     tmt_path: Optional[str] = None
diff --git a/newa/cli.py b/newa/cli.py
index 2147fef..f42dc1f 100644
--- a/newa/cli.py
+++ b/newa/cli.py
@@ -232,10 +232,9 @@ def cmd_schedule(ctx: CLIContext) -> None:
         # prepare a list of Request objects
 
         config = RecipeConfig.from_yaml_url(jira_job.recipe.url)
-        requests = config.build_requests()
 
         # create few fake Issue objects for now
-        for request in requests:
+        for request in config.build_requests():
             request_job = RequestJob(
                 event=jira_job.event,
                 erratum=jira_job.erratum,
```

PR #26, opened 2024-04-04
Implement basic `execute` subcommand

kkaarreell:
This is currently just for testing and discussion purposes.

- implements basic `execute` command running previously generated requests through Testing Farm
- adds `demodata` directory with `tmt` tree that could be used for testing


newa/__init__.py
```diff
@@ -477,9 +479,93 @@ class Request(Cloneable, Serializable):
...
+        return TFRequest(api=api, uuid=request_uuid)
+
+
+@define
+class TFRequest(Cloneable, Serializable):
+    """ A class representing plain Testing Farm request """
+
+    api: str
+    uuid: str
+    details: Optional[dict[str, Any]] = None
```
kkaarreell, line 555: @happz Would you have suggestion what should be the type here? I wanted to use `JSON` as similarly to `ErrataTool.fetch_info()` it would return JSON but when using `JSON` as type I was getting an error that it is not indexable.

newa/cli.py
```diff
@@ -231,30 +261,84 @@ def cmd_schedule(ctx: CLIContext) -> None:
...
+    #    uuid='519f5c01-46b6-47c9-a055-aecaa32e6a20')
+    print(f'{basename} TF request filed with uuid {tf_request.uuid}')
+    finished = False
+    delay = 60
+    while not finished:
+        time.sleep(delay)
+        tf_request.fetch_details()
+        state = tf_request.details['state']
+        print(f'{basename} TF reqest {tf_request.uuid} {state}')
+        finished = state == 'complete'
```
kkaarreell, line 328: Are there other state values that TF uses for completed requests? Or should I use different attribute for a test?

newa/__init__.py
```diff
@@ -477,9 +479,93 @@ class Request(Cloneable, Serializable):
...
     git_url: Optional[str] = None
     git_ref: Optional[str] = None
     tmt_path: Optional[str] = None
+    plan: Optional[str] = None
+
+    def fetch_details(self) -> None:
+        raise NotImplementedError
+
+    def generate_tf_exec_command(self) -> str:
+        cmdline_start = "NO_COLOR=yes testing-farm request --no-wait"
```
happz, line 492: I would highly recommend using a list and dictionary to construct the command line and environment variables, respectively. It's much easier to follow them in the code, for both linters and tracebacks, there is no need for placeholders like `context_params` - iterate over context, and if it's empty, nothing is added to the list. Not even the check whether `self.context` is empty is needed. And, especially, there is no need to even think about spaces used as separators - do we start each component with a space, or end with a space? Plus, `subprocess.run()` accepts command as a list too, and even has a parameter for environment variables.

```python
environment: dict[str, str] = {
    'NO_COLOR': 'yes'
}

command: list[str] = [
    'testing-farm', 'request', '--no-wait'
]

...

if self.git_ref:
    command += ['--git-ref', self.git_ref]

...

# process context
for k, v in self.context.items():
    command += ['-c', f'{k}="{v}"']

...

return command, environment


def initiate_tf_request(self) -> TFRequest:
    command, environment = self.generate_tf_exec_command()

    process = subprocess.run(
        ' '.join(command),
        env=environment,
        ...
```

newa/__init__.py
```diff
@@ -477,9 +479,93 @@ class Request(Cloneable, Serializable):
...
+        # check compose
+        if not self.compose:
+            print('ERROR: compose is not specified for the request')
```
happz, line 495: Shouldn't we raise an exception instead? It will end up that way anyway, `return ''` here leads to `raise Exception` on line 528.

newa/__init__.py
happz: Hm, I wonder where that complain comes from, because `details: Optional[JSON] = None` should be fine.

newa/__init__.py
```diff
@@ -477,9 +479,93 @@ class Request(Cloneable, Serializable):
...
+            url=self.api,
+            response_content=ResponseContentType.JSON)
+
+
+@define
+class Execution(Cloneable, Serializable):
+    """ A test job execution """
+
+    return_code: Optional[int] = None
+    artifacts: Optional[str] = None
```
happz, line 568: I'd go with `artifacts_url`, not just `artifacts` - IIUIC, this field points to them, does not contain them.

newa/cli.py
```diff
@@ -231,30 +261,87 @@ def cmd_schedule(ctx: CLIContext) -> None:
...
+    schedule_list = [
+        os.path.join(
+            ctx.state_dirpath,
+            child.name)
+        for child in ctx.state_dirpath.iterdir()
+        if child.name.startswith('schedule-')]
+
+    worker_pool = multiprocessing.Pool(workers)
+    for _ in worker_pool.map(worker, schedule_list):
+        # small sleep to avoid race conditions inside tmt code
```
happz, line 306: What race conditions do we have in tmt? :)

newa/cli.py
happz: `error` should be handled too.

newa/__init__.py
kkaarreell: Forget it. It does not complain now, I think it got somehow fixed by the most recent changes I did, which was moving some code from `cli.py` to classess in `__init__.py`.

newa/__init__.py
kkaarreell: Makes very sense

newa/cli.py
```diff
@@ -231,30 +261,87 @@ def cmd_schedule(ctx: CLIContext) -> None:
...
+    # read request details
+    schedule_job = ScheduleJob.from_yaml_file(Path(schedule_file))
+    print(f'{basename} initiating TF request')
+    tf_request = schedule_job.request.initiate_tf_request()
+    # tf_request = TFRequest(
+    #    api='https://api.dev.testing-farm.io/v0.1/requests/519f5c01-46b6-47c9-a055-aecaa32e6a20',
+    #    uuid='519f5c01-46b6-47c9-a055-aecaa32e6a20')
+    print(f'{basename} TF request filed with uuid {tf_request.uuid}')
+    finished = False
+    delay = 60
```
happz, line 325: I'd recommend making this value configurable - I know about engineers who would try to run checks more often, and I for one would be perfectly fine with checking every 10 minutes. Maybe an envvar would be easy to use, with a default?

newa/__init__.py
kkaarreell: Oh, I forgot to update this one. Will fix it.

newa/__init__.py
kkaarreell: ack

newa/cli.py
```diff
@@ -231,30 +261,87 @@ def cmd_schedule(ctx: CLIContext) -> None:
...
+        # small sleep to avoid race conditions inside tmt code
+        time.sleep(0.1)
+
+    print('Done')
+
+
+def worker(schedule_file: str) -> None:
+    workdir, basename = os.path.split(schedule_file)
+
+    print(f'{basename} processing request...')
```
happz, line 315: I think we should be able to construct a helper that would add the prefix automagically. Can you try the following?

```python
from functools import partial

log = partial(print, basename)

...

log('processing request...')

# which translates to print(basename, 'processing request...') which emits "{basename} processing request..."
```

newa/cli.py
```diff
@@ -231,30 +261,87 @@ def cmd_schedule(ctx: CLIContext) -> None:
...
+        jira=schedule_job.jira,
+        recipe=schedule_job.recipe,
+        request=schedule_job.request,
+        execution=Execution(return_code=0, artifacts=tf_request.details['run']['artifacts']),
+        )
 
-        ctx.save_erratum_job('execute-', erratum_job)
+    name = basename.replace('schedule-', 'execute-')
+    output_file = os.path.join(workdir, name)
+    execute_job.to_yaml_file(Path(output_file))
```
happz, line 343: Can be simplified, thanks to `pathlib` magic (`Path / str` yield `Path`), to:

```python
execute_job.to_yaml_file(workdir / basename.replace('schedule-', 'execute-'))
```

newa/cli.py
```diff
@@ -231,30 +261,87 @@ def cmd_schedule(ctx: CLIContext) -> None:
...
-        # run() returns result object
-        # result = worker.run()
+    if not os.environ.get("TESTING_FARM_API_TOKEN"):
+        raise ValueError("TESTING_FARM_API_TOKEN not set!")
+
+    # get a list of files to be scheduled so that they can be distributed across workers
+    schedule_list = [
+        os.path.join(
+            ctx.state_dirpath,
+            child.name)
```
happz, line 300: Please, stick to new and recommended https://docs.python.org/3/library/pathlib.html, `os.path` is supposed to be pushed out of our collective memory :)

```python
schedule_list = [
    ctx.state_dirpath / child.name
    for ...
```

newa/cli.py
```diff
@@ -231,30 +261,87 @@ def cmd_schedule(ctx: CLIContext) -> None:
...
+
+    worker_pool = multiprocessing.Pool(workers)
+    for _ in worker_pool.map(worker, schedule_list):
+        # small sleep to avoid race conditions inside tmt code
+        time.sleep(0.1)
+
+    print('Done')
+
+
+def worker(schedule_file: str) -> None:
```
happz, line 312: `Path` - wherever a path appears, it must be a `Path` instance, not `str`. It provides access to various path-related methods, and it signifies the variable is not just any string, it's a *path*.

newa/cli.py
happz: BTW a very helpful table to ease the transition: https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module

newa/cli.py
kkaarreell: Hm, now I feel dumb not filing a report right when hitting the issue in the past. And of course, I cannot reproduce it now. IIRC the issue was with starting multiple threads starting tmt in parallel and the error was probably somehow related to a collision with run directory name? But I could be wrong.

newa/cli.py
```diff
@@ -231,30 +261,87 @@ def cmd_schedule(ctx: CLIContext) -> None:
...
+    workdir, basename = os.path.split(schedule_file)
```
happz, line 313: `workdir, basename = schedule_file.parent, schedule_file.name`

And since `workdir` is used just once, and `basename` just for logging and at the end, to construct the path, are they necessary? With the `partial` trick below, we don't need `basename` to exist for logging, `partial(print, workdir.name)` would work as well, and the end might be refactored:

```python
execute_job.to_yaml_file(schedule_file.parent / schedule_file.name.replace('schedule-', 'execute-'))
```

kkaarreell:
Great tips, thank you very much!

newa/cli.py
happz: Ah, yep, might be. I'd definitely like to fix it in tmt, we were already discussing taking a lock on workdir root, creating the workdir, and releasing the lock, to avoid collisions. You might have hit that.

newa/cli.py

newa/__init__.py

newa/__init__.py

newa/cli.py

newa/cli.py
kkaarreell: updated

newa/cli.py

newa/cli.py

newa/cli.py

newa/cli.py

kkaarreell:
@happz All review comments should be addressed. Thank you.

newa/cli.py
happz: Hmm, is it actually needed? The code behind `worker` submits a job to TF, but does it run some `tmt run` commands on its own? If not, then we should be fine, because there should be absolutely no conflict between various `tmt run` commands executed by TF on behalf of different TF requests.

kkaarreell: I wanted to implement direct `tmt` execution in the future. We had it in our POC as well and it was quite handy.

happz: Understood, let's keep it then until it gets fixed in tmt.

PR #27, opened 2024-04-18
Implement basic 'report' subcommand

kkaarreell:
- Modifies `execute` command so that test results are imported to ReportPortal
 - Implements `report` command that searches for imported launches in ReportPortal and merges them to a single one and reports final RP launch back to Jira.

### DIFF
--- a/newa/__init__.py
+++ b/newa/__init__.py
@@ -588,7 +588,7 @@ class Request(Cloneable, Serializable):
                 '--tmt-environment',
                 'TMT_PLUGIN_REPORT_REPORTPORTAL_LAUNCH=ksrot_newa',
                 '--tmt-environment',
-                'TMT_PLUGIN_REPORT_REPORTPORTAL_SUITE_PER_PLAN=true',
+                'TMT_PLUGIN_REPORT_REPORTPORTAL_SUITE_PER_PLAN=1',
                 '--context', f'newa_batch={self.get_hash(ctx.timestamp)}',
                 ]
         # check compose
--- a/newa/__init__.py
+++ b/newa/__init__.py
@@ -26,6 +26,7 @@ from typing import (
     cast,
     overload,
     )
+from urllib.parse import quote as Q  # noqa: N812
 
 import attrs
 import jinja2
@@ -277,7 +278,7 @@ def eval_test(
 
 
 def get_url_basename(url: str) -> str:
-    return os.path.splitext(os.path.basename(url))[0]
+    return os.path.splitext(os.path.basename(urllib.parse.urlparse(url).path))[0]
 
 
 class EventType(Enum):
@@ -344,13 +345,13 @@ class ErrataTool:
 
     def fetch_info(self, erratum_id: str) -> JSON:
         return get_request(
-            url=f"{self.url}/advisory/{erratum_id}.json",
+            url=f"{self.url}/advisory/{Q(erratum_id)}.json",
             krb=True,
             response_content=ResponseContentType.JSON)
 
     def fetch_releases(self, erratum_id: str) -> JSON:
         return get_request(
-            url=f"{self.url}/advisory/{erratum_id}/builds.json",
+            url=f"{self.url}/advisory/{Q(erratum_id)}/builds.json",
             krb=True,
             response_content=ResponseContentType.JSON)
 
@@ -1028,13 +1029,14 @@ class ReportPortal:
     project: str
 
     def get_launch_url(self, launch_id: str) -> str:
-        return f"{self.url}/ui/#{self.project}/launches/all/{launch_id}"
+        return f"{self.url}/ui/#{Q(self.project)}/launches/all/{Q(launch_id)}"
 
     def get_request(self,
                     path: str,
                     params: Optional[dict[str, str]] = None,
                     version: int = 1) -> JSON:
-        url = urllib.parse.urljoin(self.url, f'/api/v{version}/{self.project}/{path.lstrip("/")}')
+        url = urllib.parse.urljoin(
+            self.url, f'/api/v{version}/{Q(self.project)}/{Q(path.lstrip("/"))}')
         if params:
             url = f'{url}?{urllib.parse.urlencode(params)}'
         headers = {"Authorization": f"bearer {self.token}", "Content-Type": "application/json"}
@@ -1047,7 +1049,7 @@ class ReportPortal:
                      path: str,
                      json: JSON,
                      version: int = 1) -> JSON:
-        url = f'{self.url}/api/v{version}/{self.project}/{path.lstrip("/")}'
+        url = f'{self.url}/api/v{version}/{Q(self.project)}/{Q(path.lstrip("/"))}'
         headers = {"Authorization": f"bearer {self.token}", "Content-Type": "application/json"}
         req = requests.post(url, headers=headers, json=json)
         if req.status_code == 200:
--- a/newa/__init__.py
+++ b/newa/__init__.py
@@ -785,7 +785,6 @@ class ErratumConfig(Serializable):  # type: ignore[no-untyped-def]
         factory=list, converter=lambda issues: [
             IssueAction(**issue) for issue in issues])
 
-<<<<<<< HEAD
 
 @frozen
 class IssueHandler:
@@ -1003,6 +1002,7 @@ class IssueHandler:
 #
 # ReportPortal communication
 #
+
 
 @define
 class ReportPortal:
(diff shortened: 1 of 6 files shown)


kkaarreell:
Just FYI, this PR has been rebased on top of the main branch, which includes jira integration.
Also, the current code refers to this branch in demo config files to ease testing. Before merging, it has to be replaced with a reference to the `main` branch.

newa/__init__.py
```diff
@@ -554,13 +567,29 @@ class Request(Cloneable, Serializable):
...
+        if rp_token and rp_url:
+            command += [
+                '--tmt-environment',
+                f'TMT_PLUGIN_REPORT_REPORTPORTAL_TOKEN={rp_token}',
+                '--tmt-environment',
+                f'TMT_PLUGIN_REPORT_REPORTPORTAL_URL={rp_url}',
+                '--tmt-environment',
+                'TMT_PLUGIN_REPORT_REPORTPORTAL_PROJECT=baseosqe',
+                '--tmt-environment',
+                'TMT_PLUGIN_REPORT_REPORTPORTAL_LAUNCH=ksrot_newa',
```
happz, line 588: Would it make sense to read project and launch from configuration right away? It does look like a temporary, testing set of values, I get that.

newa/__init__.py
```diff
@@ -634,8 +663,9 @@ def fetch_details(self) -> None:
 class Execution(Cloneable, Serializable):
     """ A test job execution """
 
-    return_code: Optional[int] = None
+    return_code: Optional[int] = 0
```
happz, line 666: Does this field need to be `Optional` when the default is no longer `None`? It does not seem to be used anywhere but here and at https://github.com/RedHatQE/newa/pull/27/files#diff-fbfe615aa583117b7a57302f5da6ff45c1eda26a6f4b2406b68a09ffe1bd698aR343, where the value is even explicitly given. Does the field need to be `Optional` or even have a default value? It seems to be set to `0` right from the beginning when instance of `Execution` is created :)

The same seems to apply to `batch_id` - it has default value, but that value is never used. When `Execution` is instantiated, `batch_id` is set at that moment, https://github.com/RedHatQE/newa/pull/27/files#diff-fbfe615aa583117b7a57302f5da6ff45c1eda26a6f4b2406b68a09ffe1bd698aR345.

Both fields seem to be initialized by whoever creates an instance, therefore it would be safer to drop the default values. Then we can catch possible call sites where `Execution` is created without providing these values.

newa/__init__.py
```diff
@@ -968,3 +998,194 @@ def drop_obsoleted_issue(self, issue: Issue, obsoleted_by: Issue) -> None:
...
+    def get_launch_url(self, launch_id: str) -> str:
+        return f"{self.url}/ui/#{self.project}/launches/all/{launch_id}"
+
+    def get_request(self,
+                    path: str,
+                    params: Optional[dict[str, str]] = None,
+                    version: int = 1) -> JSON:
+        base_url = f'{self.url}/api/v{version}/{self.project}/{path.lstrip("/")}'
+        get_params = '&'.join([f'{k}={v}' for (k, v) in params.items()]) if params else ''
+        url = f'{base_url}?{get_params}'
```
happz, line 1023: I would strongly recommend using https://docs.python.org/3/library/urllib.parse.html instead of manual processing of URL components. Namely `urlencode` and `urljoin`, take care of the right number of `/` and quoting all parameter names and values. Something like this should give you the right answer:

```python
url = urllib.parse.urljoin(self.url, f'/api/v{version}/{self.project}/{path.lstrip("/")}')

if params:
  url = f'{url}?{urllib.parse.urlencode(params)}'
```

newa/__init__.py
```diff
@@ -968,3 +998,194 @@ def drop_obsoleted_issue(self, issue: Issue, obsoleted_by: Issue) -> None:
...
+        if attributes:
+            for key, value in attributes.items():
+                query_data['attributes'].append({"key": key.strip(), "value": value.strip()})
+        print(f'Merging launches: {launch_ids}')
+        data = self.post_request('/launch/merge', json=query_data)
+        if data:
+            print('Launches successfully merged')
+            return str(data['id'])
+        print('Failed to merge launches')
+        return ''
```
happz, line 1073: It's safer to use `None` than an empty string for error-like return values. It's then easier for linters to follow them around. Consider how this return value is used:

```python
            if len(to_merge) > 1:
                final_launch = rp.merge_launches(
                    to_merge, jira_launch_name_mapping[jira_id], description, {})
            else:
                final_launch = to_merge[0]
            launch_url = rp.get_launch_url(final_launch)
```

If `merge_launches()` fails to merge launches, it'd return an empty string which would then be passed to `get_launch_url()`, leading to "launch URL" `{self.url}/ui/#{self.project}/launches/all/`. I don't know whether such an URL is expected, or whether an empty string is propagated on purpose, but that's the point: it's a tricky value, one that looks like an error is being reported, then it's used like nothing happened. If `None` would be returned instead, there would have to be special handling for `final_launch` being `None` - is it an error, or do we get URL that lists all launches, and so on. An empty string can easily spread doubt through the code.

newa/cli.py
```diff
@@ -388,12 +285,14 @@ def cmd_schedule(ctx: CLIContext) -> None:
 @main.command(name='execute')
 @click.option(
     '--workers',
-    default=4,
+    default=8,
```
happz, line 288: I remember our discussion when this default was introduced. Isn't it time to make it something that wouldn't have to be tweaked in the source, e.g. (CPU count / 2), or derive the default from some environment variable?

kkaarreell: I would really prefer to do this later. IMHO it doesn't make sense to set any limits based on CPU count since those processes are mostly waiting anyway. Moreover, a user can run multiple `newa` instances in parallel anyway.

newa/__init__.py
kkaarreell: `rp_project` is already read from a config file, clearly I forgot to update it here.
Regarding `rp_launch`, I have decided to add it to the `Request` class and read its value from the recipe of or (if not defined) use recipe file basename as a launch name. I think it makes more sense from a user perspective, although we will most likely change it in the future anyway.
Both should be addressed now.

newa/__init__.py
kkaarreell: Fixed.

newa/__init__.py
kkaarreell: Thank you. Fixed.

newa/__init__.py
kkaarreell: Fixed. I have rewritten launch merging a bit.

newa/__init__.py
```diff
@@ -273,6 +276,10 @@ def _test_match(s: str, pattern: str) -> bool:
     return bool(outcome == 'true')
 
 
+def get_url_basename(url: str) -> str:
+    return os.path.splitext(os.path.basename(url))[0]
```
happz, line 280: I would recommend using `urllib.parse.urlparse` instead, see https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse

newa/__init__.py
```diff
@@ -968,3 +1014,194 @@ def drop_obsoleted_issue(self, issue: Issue, obsoleted_by: Issue) -> None:
...
+
+@define
+class ReportPortal:
+
+    token: str
+    url: str
+    project: str
+
+    def get_launch_url(self, launch_id: str) -> str:
+        return f"{self.url}/ui/#{self.project}/launches/all/{launch_id}"
```
happz, line 1031: To stay on the safe side, I strongly suggest using `urllib.parse.quote` (https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote) when injecting our strings into URLs. Our strings may contain characters that must not appear in URL - space, maybe even diacritics, or characters like `?`. I suppose most of the time, this will not be the case, but it's a good and safe habit.

```python
from urllib.parse import quote as Q

return f"{self.url}/ui/#{Q(self.project)}/launches/all/{Q(launch_id)}"
```

It should be applied to pretty much every component we use to construct URL, except the schema (`urllib.parse.quote('http://foo.bar.com')` => `'http%3A//foo.bar.com'`).

newa/__init__.py

newa/__init__.py

newa/__init__.py
```diff
@@ -273,6 +277,10 @@ def _test_match(s: str, pattern: str) -> bool:
     return bool(outcome == 'true')
 
 
+def get_url_basename(url: str) -> str:
+    return os.path.splitext(os.path.basename(urllib.parse.urlparse(url).path))[0]
```
happz, line 281: Is `splitext` still needed?

kkaarreell: basename will provide filename including an extension but I want to cut the extension off. I can see the function name is misleading in this context. Since I am going to use the final name as a default name for RP launch, I think I will move the `splitext` call to the respective place and let this function to return truly a basename. Will fix it right away.

happz (approved):

PR #19, opened 2024-03-25
Bootstrap test evaluation

happz:
Related to https://github.com/RedHatQE/newa/issues/8

PR #29, opened 2024-04-19
Enable condition evaluation for jira issue config and job recipe

kkaarreell:
- updates demodata
- enables 'when' evaluation for Jira issue config
- enables 'when' evaluation for job recipe
- updates `_merge_key()`  to be able to merge lists

### DIFF
--- a/demodata/recipe.yaml
+++ b/demodata/recipe.yaml
@@ -12,8 +12,10 @@ dimensions:
         CITY: Brno
     - environment:
         CITY: Boston
+      when: CONTEXT.color is match("blue")
   colors:
     - context:
         color: red
+      when: ENVIRONMENT.CITY is not match("Brno")
     - context:
         color: blue
--- a/newa/__init__.py
+++ b/newa/__init__.py
@@ -449,9 +449,10 @@ class RawRecipeConfigDimension(TypedDict, total=False):
     compose: Optional[str]
     git_url: Optional[str]
     git_ref: Optional[str]
+    when: Optional[str]
 
 
-_RecipeConfigDimensionKey = Literal['context', 'environment', 'git_url', 'git_ref']
+_RecipeConfigDimensionKey = Literal['context', 'environment', 'git_url', 'git_ref', 'when']
 
 
 # A list of recipe config dimensions, as stored in a recipe config file.
@@ -489,13 +490,21 @@ class RecipeConfig(Cloneable, Serializable):
                 dest: RawRecipeConfigDimension,
                 src: RawRecipeConfigDimension,
                 key: str) -> None:
-            if key not in dest:
+            # instruct how individual attributes should be merged
+            # attribute 'when' needs special treatment as we are joining conditions with 'and'
+            if key == 'when' and ("when" not in dest) and src["when"]:
+                dest['when'] = f'( {src["when"]} )'
+            elif key == 'when' and dest["when"] and src["when"]:
+                dest['when'] += f' and ( {src["when"]} )'
+            elif key not in dest:
                 # we need to do a deep copy so we won't corrupt the original data
                 dest[key] = copy.deepcopy(src[key])  # type: ignore[literal-required]
             elif isinstance(dest[key], dict) and isinstance(src[key], dict):  # type: ignore[literal-required]
                 dest[key].update(src[key])  # type: ignore[literal-required]
+            elif isinstance(dest[key], list) and isinstance(src[key], list):  # type: ignore[literal-required]
+                dest[key].extend(src[key])  # type: ignore[literal-required]
             else:
-                raise Exception(f"Don't know how to merge record type {key}")
+                raise Exception(f"Don't know how to merge record type '{key}'")
 
         def merge_combination_data(
                 combination: tuple[RawRecipeConfigDimension, ...]) -> RawRecipeConfigDimension:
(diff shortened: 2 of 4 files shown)


newa/__init__.py
```diff
@@ -677,10 +703,25 @@ class IssueAction:  # type: ignore[no-untyped-def]
...
     issues: list[IssueAction] = field(  # type: ignore[var-annotated]
         factory=list, converter=lambda issues: [
             IssueAction(**issue) for issue in issues])
+
+
+#
+# Recipe variables
+#
+
+class RecipeGenVar:
```
kkaarreell, line 720: Here I really don't know how to approach a class which may have "arbitrary" attributes.

demodata/recipe.yaml
```diff
@@ -12,8 +12,12 @@ dimensions:
         CITY: Brno
     - environment:
         CITY: Boston
+      when:
+        - CONTEXT.color is match("blue")
   colors:
     - context:
         color: red
+      when:
+        - JOB.environment.CITY is not match("Brno")
```
kkaarreell, line 21: While for Jira issue config `when` is a string, here it seems more practical to have a list, since we are merging it. However, all conditions in such a list must match (AND), while users may intuitively understand it as any condition would be sufficient. WDYT? Should we stick to a single condition (string) and somehow rewrite the merging or should we rather support OR within a single `when` (before merging)?

happz: From Ansible we're used to treating the list of conditions as if an implicit `and` is applied to all items. `or` in a single condition should already be supported, together with `and` and brackets. I did not check for this, but even multiline conditions might work (needs to be tested), e.g.

```yaml
when: >-
  EVENT.type is erratum
  and (EVENT.type != bodhi or EVENT.id is not match("1.*"))
```

I would probably stick to string only, i.e. no lists, but all forms would work.

kkaarreell: @happz I have pushed an update, using a string for "when" conditions. However, I am facing some pre-commit issues I am not able to deal with.
{{{
newa/__init__.py:459: error: Unsupported left operand type for + ("None")  [operator]
newa/__init__.py:459: note: Left operand is of type "Optional[str]"
}}}
Really not sure how to properly approach this one.
{{{
newa/__init__.py:503: error: Extra argument "when" from **args for "Request"  [misc]
}}}
I though having "when" for a Request class is not necessary, however linter thinks we are passing this atribute as well.

happz:
```
newa/__init__.py:459: error: Unsupported left operand type for + ("None") [operator]
newa/__init__.py:459: note: Left operand is of type "Optional[str]"

            elif key == 'when' and src["when"]:
                dest['when'] += f' and ( {src["when"]} )'
```

The warning is about the left side of the assignment, `dest['when']`. We know that `key` is `when`, and that `src['when']` is set and is true-ish, but `dest['when']` is still `Optional[str]`. Thanks to the previous branch we know it's set, but that's all, you still need to deal with `dest['when'] == None` or `dest['when'] == ''`. And `None + str` is what the complaint is about. Adding `dest['when']` check & handling for these would resolve the issue.

```
newa/init.py:503: error: Extra argument "when" from **args for "Request" [misc]
```

Correct. `**combination` - `combination` is of type `RawRecipeConfigDimension` and as such may contain the `when` key. You either need to teach `Request` to accept it, or, if that's not the right thing to do, drop it from `combination` before passing it to `Request`.

kkaarreell: Dropping it is what I am trying to do with `del combination["when"]` but it doesn't seem to work.

happz: Would it get better if you moved it one level to the left? I suspect it's hard for a linter to prove `when` won't be there, e.g. there's `if not test_result:` which may lead to `when` being still present.

newa/__init__.py
happz: I'm not sure it's necessary to have one, I think we can easily use `RecipeContext` and `RecipeEnvironment`. Then, it also seems to me `RecipeJobVar` is unnecessary too, it seems to me the code around lines 484 can be simplified.

Definitely `variables.context` and `variables.environment` are basically just copying keys/values from a dictionary to a container class, `ENVIRONMENT=combination.get('environment', {})` and `CONTEXT=combination.get('context', {})` should work perfectly fine. I think this is caused by you not being aware of one trick Jinja supports in its templates, dictionary values can be retrieved by using dotted notation:

```
foo:
  bar: baz

# these are equivalent:
{{ foo["bar"] }}
{{ foo.bar }}
```

This means `CONTEXT` and `ENVIRONMENT` can be dictionaries, and still one would be able to access their content using dots, e.g. `{{ ENVIRONMENT.USER }}` would be translated to `ENVIRONMENT["USER"]` under the hood. `ENVIRONMENT` itself does not have to be a class, and we can re-use existing dictionaries.

What I don't understand is the necessity of `JOB` variable, would it make sense to drop it & use `COMPOSE=combination['compose']` instead? We already expose its `environment` and `context` attributes as standalone variables `ENVIRONMENT` and `CONTEXT`. Will there be more to store in `RecipeJobVar`?

kkaarreell: > I think this is caused by you not being aware of one trick Jinja supports in its templates, dictionary values can be retrieved by using dotted notation

You are right.

> What I don't understand is the necessity of `JOB` variable
It is not necessary. I just tried to follow your pattern in https://github.com/RedHatQE/newa/pull/19/files where `JOB` could be (probably incorrectly?) understood as a common parent from where one can access available parameters.

happz: I see. That one was mostly an example/test case for the duration of bootstrapping the needed code, demonstrating the possibilities, i.e. as an option how we may expose some attributes as standalone variables if needed, to simplify the conditions. I wouldn't say a container like that has to exist - it may, if we have an object to expose at hand already, but I wouldn't see it as a mandatory requirement. Sorry for the confusion.

kkaarreell: OK. Let's get rid of `JOB` for now but in the future I would like to have something like that. Let's discuss it on a mtg.

kkaarreell: Updated.

newa/__init__.py
```diff
@@ -484,6 +504,8 @@ class Request(Cloneable, Serializable):
     git_ref: Optional[str] = None
     tmt_path: Optional[str] = None
     plan: Optional[str] = None
+    # 'when' not really needed, addit it to silent the linter
```
happz, line 507:
```suggestion
    # TODO: 'when' not really needed, adding it to silent the linter
```

The-Mule:
LGTM

PR #30, opened 2024-04-19
Read settings from newa config file

kkaarreell:
- Add support for loading settings from a newa config file
- move method for fetching Errata objects from Erratum class to ErrataTool class

### DIFF
--- a/newa/__init__.py
+++ b/newa/__init__.py
@@ -8,6 +8,7 @@ import re
 import subprocess
 import time
 from collections.abc import Iterator
+from configparser import ConfigParser
 from enum import Enum
 from pathlib import Path
 from typing import (
@@ -42,6 +43,7 @@ if TYPE_CHECKING:
 
 T = TypeVar('T')
 SerializableT = TypeVar('SerializableT', bound='Serializable')
+SettingsT = TypeVar('SettingsT', bound='Settings')
 
 
 def yaml_parser() -> ruamel.yaml.YAML:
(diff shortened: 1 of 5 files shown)


newa/__init__.py
```diff
@@ -106,6 +108,52 @@ def render_template(
...
+
+    @classmethod
+    def load(cls: type[SettingsT], config_file: Path) -> Settings:
+        cp = ConfigParser()
+        cp.read(config_file)
+
+        def __get(
+                cp: ConfigParser,
+                path: str,
+                envvar: Optional[str] = None,
```
happz, line 136: It seems that all `__get()` calls specify `envvar`, `str` and no default value would reflect the situation better.

newa/__init__.py
```diff
@@ -106,6 +108,52 @@ def render_template(
...
+    rp_url: Optional[str] = None
+    rp_token: Optional[str] = None
+    rp_project: Optional[str] = None
+    jira_url: Optional[str] = None
+    jira_token: Optional[str] = None
+    jira_project: Optional[str] = None
+    tf_token: Optional[str] = None
+    tf_recheck_delay: Optional[str] = None
+
+    def get(self, key: str, default: Optional[str] = '') -> str:
```
happz, line 125: Can the `default` ever be `None`? Even the default value is a string. If it can be `None`, then there is a surprise waiting in the code below: `settings.get('foo', default=None) -> getattr(self, 'foo', None) -> str(None) -> get() returns "None"`. I'd recommend dropping either `Optional`, or `str`, either one should go: `Optional` does not seem to apply in any call, and if it does and `None` should be allowed, `str()` will mask invalid keys, providing bogus return values for them.

newa/cli.py
```diff
@@ -291,8 +302,12 @@ def cmd_schedule(ctx: CLIContext) -> None:
 def cmd_execute(ctx: CLIContext, workers: int) -> None:
     ctx.enter_command('execute')
 
-    if not os.environ.get("TESTING_FARM_API_TOKEN"):
+    tf_token = ctx.settings.get("tf_token")
```
happz, line 305: Same as above, `ctx.settings.tf_token` should work and be verifiable by linters.

newa/cli.py
```diff
@@ -291,8 +302,12 @@ def cmd_schedule(ctx: CLIContext) -> None:
...
+    if not tf_token:
         raise ValueError("TESTING_FARM_API_TOKEN not set!")
+    # make some variables available for workers
+    os.environ["TESTING_FARM_API_TOKEN"] = tf_token
+    os.environ["NEWA_TF_RECHECK_DELAY"] = ctx.settings.get("tf_recheck_delay")
```
happz, line 310: This does not look right. Code takes values from loades settings, puts them into environment, and the worker code then ignores settings and peeks into environment for *some* keys.

Would you mind checking `Pool.starmap()` instead of `Pool.map()`? That should let us pass more arguments to `worker()`, and we could propagate `ctx.settings`. Something like this should work:

```python
schedule_list = [
  (ctx, ctx.state_dirpath / child.name)
  for child in ctx.state_dirpath.iterdir()
  if child.name.startswith('schedule-')]

...

for _ in worker_pool.starmap(worker, schedule_list):

...

def worker(ctx: CLIContext, schedule_file: Path) -> None:
    ...
    delay = int(ctx.settings.tf_recheck_delay)
```

newa/__init__.py
```diff
@@ -106,6 +108,52 @@ def render_template(
...
+        return str(getattr(self, key, default))
+
+    @classmethod
+    def load(cls: type[SettingsT], config_file: Path) -> Settings:
+        cp = ConfigParser()
+        cp.read(config_file)
+
+        def __get(
```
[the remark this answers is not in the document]
kkaarreell: Fixed

newa/__init__.py
kkaarreell: I thought there might be options where envvar override won't be available but currently there aren't any and maybe there won't be ever. I have changed it.

newa/__init__.py
kkaarreell: I must admit that my understanding when `Optional` should be used was wrong. I have dropped it from the code on few other places and well.

newa/cli.py
```diff
@@ -173,11 +180,15 @@ def cmd_event(ctx: CLIContext, errata_ids: list[str]) -> None:
     if not errata_ids:
         raise Exception('Missing errata IDs!')
 
+    et_url = ctx.settings.get('et_url')
```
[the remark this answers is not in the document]
kkaarreell: Right.

newa/cli.py
kkaarreell: Fixed.

newa/cli.py
kkaarreell: Fixed. Thank you, I was not aware of `starmap`.

happz (approved):

PR #34, opened 2024-05-31
Add container erratum support

kkaarreell:
Adds support for container type advisories.
Adds `action.newa_id` parameter allowing a user to override NEWA issue ID in the jira issue config file
Adds `components` attribute to Erratum class
Adds demodata/jira-container-config.yaml presenting possible usage

### DIFF
--- a/newa/__init__.py
+++ b/newa/__init__.py
@@ -73,6 +73,7 @@ def yaml_parser() -> ruamel.yaml.YAML:
         return representer.represent_scalar('tag:yaml.org,2002:str', data.value)
 
     yaml.representer.add_representer(EventType, _represent_enum)
+    yaml.representer.add_representer(ErratumContentType, _represent_enum)
 
     return yaml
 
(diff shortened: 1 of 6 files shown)


newa/__init__.py
```diff
@@ -407,12 +408,18 @@ def get_errata(self, event: Event) -> list[Erratum]:
...
-                                      respin_count=int(info_json["respin_count"]),
-                                      summary=info_json["synopsis"],
-                                      people_assigned_to=info_json["people"]["assigned_to"],
-                                      release=release,
-                                      builds=builds))
+                errata.append(
+                    Erratum(
+                        id=event.id,
+                        content_type=ErratumContentType(
+                            info_json["content_types"][0]),
```
The-Mule, line 415: Can we a check that there are some content_types? Or do you think it is impossible that at some point that key can have empty list?

newa/__init__.py
```diff
@@ -730,6 +747,18 @@ def id(self) -> str:
         raise NotImplementedError
 
 
+class NVRParser:
```
The-Mule, line 750: Maybe this need not to be a class? Do we need it for something else than to get name from a build? If no then I suggest to have a method for that only.

newa/__init__.py
kkaarreell: TBH I have no idea if there can be erratum without a content type. But what about something like this?
```
if not info_json["content_types"]:
    warn('Erratum XX does not have a content type)
    content_type = None
else:
    content_type = ErratumContentType(info_json["content_types"][0])
...
...
Erratum(
                        id=event.id,
                        content_type=content_type
...

```

newa/__init__.py
kkaarreell: In fact, I was thinking there could be some methods in the future (maybe some NVR arithmetics etc.), just don't have a use case for it right now. 
Right now we need it for name but at the same time this class provides also version and release.
I think I like the object approach a bit more.
@happz Opinions?

newa/__init__.py
The-Mule: It is just was nitpick, info_json["people"]["assigned_to"] is not handled too so maybe we can just leave it as it is (if there is no list there it will raise an Exception anyway).

The-Mule (approved):
All in all, it looks good, I did not spot any real issue.

newa/__init__.py
kkaarreell: OK. Let's leave it as it is then. :-)

newa/__init__.py
kkaarreell: Done.

happz (approved):

PR #36, opened 2024-06-04
Separate recipe reportportal configuration

kkaarreell:
Isolate reportportal configuration into separate branch in the newa recipe file.
Additionally, utilize Jinja template rendering for these newly added attributes.

### DIFF
--- a/newa/__init__.py
+++ b/newa/__init__.py
@@ -484,18 +484,28 @@ RecipeContext = dict[str, str]
 RecipeEnvironment = dict[str, str]
 
 
+class RawRecipeReportPortalConfigDimension(TypedDict, total=False):
+    launch_name: Optional[str]
+    launch_description: Optional[str]
+    suite_description: Optional[str]
+
+
+_RecipeReportPortalConfigDimensionKey = Literal['launch_name',
+                                                'launch_description', 'suite_description']
+
+
 class RawRecipeConfigDimension(TypedDict, total=False):
     context: RecipeContext
     environment: RecipeEnvironment
     compose: Optional[str]
     git_url: Optional[str]
     git_ref: Optional[str]
-    rp_launch: Optional[str]
+    reportportal: Optional[RawRecipeReportPortalConfigDimension]
     when: Optional[str]
 
 
 _RecipeConfigDimensionKey = Literal['context',
-                                    'environment', 'git_url', 'git_ref', 'rp_launch', 'when']
+                                    'environment', 'git_url', 'git_ref', 'reportportal', 'when']
 
 
 # A list of recipe config dimensions, as stored in a recipe config file.
@@ -591,7 +601,7 @@ class Request(Cloneable, Serializable):
     compose: Optional[str] = None
     git_url: Optional[str] = None
     git_ref: Optional[str] = None
-    rp_launch: Optional[str] = None
+    reportportal: Optional[RawRecipeReportPortalConfigDimension] = None
     tmt_path: Optional[str] = None
     plan: Optional[str] = None
     # TODO: 'when' not really needed, adding it to silent the linter
@@ -616,8 +626,8 @@ class Request(Cloneable, Serializable):
             raise Exception('ERROR: ReportPortal URL is not set')
         if not rp_project:
             raise Exception('ERROR: ReportPortal project is not set')
-        if not self.rp_launch:
-            raise Exception('ERROR: ReportPortal launch is not specified')
+        if (not self.reportportal) or (not self.reportportal['launch_name']):
+            raise Exception('ERROR: ReportPortal launch name is not specified')
         command += [
             '--tmt-environment',
             f'TMT_PLUGIN_REPORT_REPORTPORTAL_TOKEN={rp_token}',
@@ -626,7 +636,7 @@ class Request(Cloneable, Serializable):
             '--tmt-environment',
             f'TMT_PLUGIN_REPORT_REPORTPORTAL_PROJECT={rp_project}',
             '--tmt-environment',
-            f'TMT_PLUGIN_REPORT_REPORTPORTAL_LAUNCH={self.rp_launch}',
+            f'TMT_PLUGIN_REPORT_REPORTPORTAL_LAUNCH={self.reportportal["launch_name"]}',
             '--tmt-environment',
             'TMT_PLUGIN_REPORT_REPORTPORTAL_SUITE_PER_PLAN=1',
             '--context', f'newa_batch={self.get_hash(ctx.timestamp)}',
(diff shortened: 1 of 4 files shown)


kkaarreell:
@happz This is just the beginning, I am rewriting `reportportal` related configuration for now just to double-check the approach. What strikes me now is that sometimes I have to access `recipe.fixtures.reportportal` as an attribute and sometimes as a dictionary key. And items below it are also dictionary keys. Moreover, I am getting following pre-commit error:
```
newa/cli.py:302: error: Item "None" of "Optional[RawRecipeReportPortalConfigDimension]" has no attribute "get"  [union-attr]
newa/cli.py:303: error: Unsupported target for indexed assignment ("Optional[RawRecipeReportPortalConfigDimension]")  [index]
```
How would you propose to implement it?

kkaarreell:
So I was able to silence the linter, basically by repeating the condition unnecessarily (unless I am missing something).
https://github.com/RedHatQE/newa/pull/36/files#diff-fbfe615aa583117b7a57302f5da6ff45c1eda26a6f4b2406b68a09ffe1bd698aR306

happz:
> So I was able to silence the linter, basically by repeating the condition unnecessarily (unless I am missing something). https://github.com/RedHatQE/newa/pull/36/files#diff-fbfe615aa583117b7a57302f5da6ff45c1eda26a6f4b2406b68a09ffe1bd698aR306

I'm not sure you can do better. Yes, you assign a valid value to `config.fixtures['reportportal'] = RawRecipeReportPortalConfigDimension()`, but that does not prove anything to mypy. Who knows what happens inside that dictionary... Those linters have some blind spots, to narrow the type, only a read works.

You may simplify with extra variables, that might work, something like this:

```python
# if RP launch name is not specified in the recipe, set it based on the recipe filename
if not config.fixtures.get('reportportal', None):
    config.fixtures['reportportal'] = reportportal_fixture = RawRecipeReportPortalConfigDimension()
```

Here I think `reportportal_fixture` should be recognized as not being `None`, next part could use it:

```python
# Populate default for config.fixtures['reportportal']['launch_name']
if not reportportal_fixture.get('launch_name', None):
    reportportal_fixture['launch_name'] = os.path.splitext(get_url_basename(jira_job.recipe.url))[0]
```

kkaarreell:
> You may simplify with extra variables, that might work, something like this:
> 
> ```python
> # if RP launch name is not specified in the recipe, set it based on the recipe filename
> if not config.fixtures.get('reportportal', None):
>     config.fixtures['reportportal'] = reportportal_fixture = RawRecipeReportPortalConfigDimension()
> ```
> 
> Here I think `reportportal_fixture` should be recognized as not being `None`, next part could use it:

I tried it but it didn't help. I think I will stick with the current extra condition. :-(

kkaarreell:
@happz would you have a "nice" way for iterating across TypeDict keys so that linter won't complain?
```
newa/cli.py:320: error: TypedDict key must be a string literal; expected one of ("launch_name", "launch_description", "suite_description")  [literal-required]
```

happz:
> @happz would you have a "nice" way for iterating across TypeDict keys so that linter won't complain?
> 
> ```
> newa/cli.py:320: error: TypedDict key must be a string literal; expected one of ("launch_name", "launch_description", "suite_description")  [literal-required]
> ```

No, the best option is silencing the linter, `# type: ignore[literal-required]`.

kkaarreell:
I believe this is ready for review. To ease the testing I have redirected demo recipe URLs to this branch, before merge I will point it to main.

newa/__init__.py
```diff
@@ -484,18 +484,28 @@ class Recipe(Cloneable, Serializable):
 RecipeEnvironment = dict[str, str]
 
 
+class RawRecipeReportPortalConfigDimension(TypedDict, total=False):
```
The-Mule, line 487: Ugh, that is quite a long name :). Why it is called raw? Can't we have it defined in a recipe class so that we can simply call it is ReportPortalConfg?

newa/__init__.py
```diff
@@ -484,18 +484,28 @@ class Recipe(Cloneable, Serializable):
...
+    launch_name: Optional[str]
+    launch_description: Optional[str]
+    suite_description: Optional[str]
+
+
+_RecipeReportPortalConfigDimensionKey = Literal['launch_name',
+                                                'launch_description', 'suite_description']
+
+
 class RawRecipeConfigDimension(TypedDict, total=False):
```
The-Mule, line 497: Same here, can't we have it inside recipe class?

newa/__init__.py
kkaarreell: It is being used in Recipe and  RawRecipeConfigDimension classes. Maybe we could define it under the Recipe class, TBH I don't know what's better and will follow other's recommendations. 
@happz Opinions?

kkaarreell: Raw prefix originates from Milos's code so I will leave the answer on him.

happz: "Raw" as in "user or service input". Often this input is then processed and transformed into better-fitting data structures.

There's no perfect solution... You want the various mappings and lists entering from YAML files to be described by typed dicts and lists and properly annotated, so the code would know that this particular mapping has these keys, and avoid typos or treating a list as a dictionary. But there's a price, as mappings and lists get nested deeper in each other, components like this one tend to get longer names, expressing the hierarchy, "a dimension of report portal configuration in a recipe"...

What I would do: I'd add a comment, or adopt a naming pattern, like `Raw_Recipe_ReportPortal_Config_Dimension`, to separate levels, expressing the hierarchy, and check it has the right fields and it's at the right place. You may also decide to drop all these `TypedDict`, but then you risk letting the structure to grow out of control.

I don't like relying on nested dictionaries and lists exactly because of this: it's so easy to start with, one can prototype quickly and ship something working. It's so tempting, it's easy to have seven levels of dictionaries in a YAML file. But then RFEs start coming, and the dictionaries start getting new keys and nest even more, and suddenly, a list appears where I expected a string and everything blows out somewhere in production. New contributors start adding new things, and this structure grows, unguarded, and wild... There's no validation. Using `TypedDict` to define partial types and then compose bigger structures from them is the only defense I know, plus a schema-based validation. Or defining proper data classes, and use `attr` to load data from YAML and materialize it as a tree of genuine Python objects rather than raw dictionaries.

kkaarreell: Hi @The-Mule ,  given the options above, what would you prefer?

The-Mule: Thanks @happz for clarification. It makes sense, alright I suppose we can leave it as it is, no need to rename it.

The-Mule (approved):

happz (approved):

PR #39, opened 2024-06-12
Introduce tmt section in the recipe YAML

kkaarreell:


demodata/recipe1.yaml
```diff
@@ -2,10 +2,11 @@ fixtures:
   environment:
     PLANET: Earth
 
-  git_url: https://github.com/RedHatQE/newa.git
-  git_ref: main
-  tmt_path: demodata
-  plan: /plan1
+  tmt:
+    url: https://github.com/RedHatQE/newa.git
+    ref: main
```
The-Mule, line 7: I am wondering if we should not rather keep follow cli paramets naming (ie. {url,ref} -> git_{url,ref}). If would safe one "mind-step" from the recipe to cli parameters.

newa/__init__.py
```diff
@@ -662,19 +668,19 @@ def generate_tf_exec_command(self, ctx: CLIContext) -> tuple[list[str], dict[str
...
+        if not self.tmt.get("url", None):
+            raise Exception('ERROR: tmt "url" is not specified for the request')
+        if self.tmt['url']:
+            command += ['--git-url', self.tmt['url']]
+        if self.tmt.get("ref") and self.tmt['ref']:
+            command += ['--git-ref', self.tmt['ref']]
+        if self.tmt.get("path") and self.tmt['path']:
+            command += ['--path', self.tmt['path']]
+        if self.tmt.get("plan") and self.tmt['plan']:
+            command += ['--plan', self.tmt['plan']]
```
The-Mule, line 683: See also my note above. Once we have it, we can simply do

`command += [f"--{k}={v}" for k, v in self.tmt.items()]`

instead of lines 676-683 with a benefit to be able to pass any TF cli parameters.

The-Mule: I would need to be to add e.g. `pipeline-type` for multihost part of the recipe.

demodata/recipe1.yaml
kkaarreell: I am still trying to follow idea that one day newa could run TMT command instead od using Testing Farm and from this perspective I find these params a bit more generic.. Also translation to TF API would look differently I think. But I May be convinced doing otherwise.

newa/__init__.py
kkaarreell: `pipeline-type` would go to the `testingfarm` section, either as a separate key or under `cli_args`. ToBeDiscussed.

demodata/recipe1.yaml
The-Mule: Discussed in person. Testing-farm API actually contains tmt sections with tmt options and only translates things like --git-url or --git-ref to their tmt equivalents. Hence I am really happy with the current implementation in the end. Sorry for false alarms :).

The-Mule (approved):
All looks good to me, thanks!

PR #40, opened 2024-06-13
Support multiple system architectures

kkaarreell:
Populates `Erratum.archs` with a list of relevant architectures obtained from the erratum. Also, enables to override architectures through the `--arch` option of the `schedule` command.
Subsequently, updates `dimensions` configuration with an extra `arch` dimension that results in creating a single`Request()` per architecture.

### DIFF
--- a/newa/__init__.py
+++ b/newa/__init__.py
@@ -77,6 +77,7 @@ def yaml_parser() -> ruamel.yaml.YAML:
 
     yaml.representer.add_representer(EventType, _represent_enum)
     yaml.representer.add_representer(ErratumContentType, _represent_enum)
+    yaml.representer.add_representer(Arch, _represent_enum)
 
     return yaml
 
@@ -304,6 +305,29 @@ class EventType(Enum):
     COMPOSE = 'compose'
 
 
+class Arch(Enum):
+    """ Available system architectures """
+
+    X86_64 = 'x86_64'
+    AARCH64 = 'aarch64'
+    S390X = 's390x'
+    PPC64LE = 'ppc64le'
+    MULTI = 'multi'
+
+    @classmethod
+    def architectures(cls: type[Arch],
+                      preset: Optional[list[Arch]] = None) -> list[Arch]:
+
+        _all = [a.value for a in Arch.__members__.values() if a != Arch.MULTI]
+
+        if not preset:
+            return [Arch('x86_64')]
+        # 'multi' is given for container advisories
+        if Arch('multi') in preset:
+            return _all
+        return list(set(_all).intersection(set(preset)))
+
+
 @define
 class Cloneable:
     """ A class whose instances can be cloned """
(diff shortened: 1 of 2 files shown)


kkaarreell:
I would appreciate advise on how to represent individual architectures in the code.

newa/__init__.py
```diff
@@ -458,6 +479,21 @@ class ErratumContentType(Enum):
     DOCKER = 'docker'
 
 
+def get_testing_architectures(preset: Optional[list[SysArchType]] = None) -> list[SysArchType]:
```
kkaarreell, line 482: Is it OK to have such a function here or would you propose a different approach?

kkaarreell: I have moved it as a classmethod under `SysArchType` class.

newa/cli.py
```diff
@@ -311,9 +316,20 @@ def cmd_schedule(ctx: CLIContext) -> None:
 
         # would it be OK not to pass compose to TF? I guess so
         compose = jira_job.compose.id if jira_job.compose else None
+        if arch:
+            architectures = SysArchType.architectures(
+                [SysArchType(a) for a in arch.strip().split(',')])
```
happz, line 321: `a.strip()` otherwise this would blow out meeting `aarch64, x86_64`.

newa/__init__.py
```diff
@@ -301,6 +302,33 @@ class EventType(Enum):
     COMPOSE = 'compose'
 
 
+class SysArchType(Enum):
```
happz, line 305: I'd call it just `Arch`, because architecture is a very much established concept, and there's hardly to be "non-sys" arch.

newa/__init__.py
```diff
@@ -301,6 +302,33 @@ class EventType(Enum):
...
+
+    @classmethod
+    def architectures(cls: type[SysArchType],
+                      preset: Optional[list[SysArchType]] = None) -> list[SysArchType]:
+
+        _all = [
+            SysArchType('x86_64'),
+            SysArchType('s390x'),
+            SysArchType('aarch64'),
+            SysArchType('ppc64le')]
```
happz, line 322: To avoid the need to update this list when new architecture appears (and forgetting to do so...) you can use `[a.value for a in SysArchType.__members__.values() if a != SysArchType.MULTI]`

newa/cli.py
```diff
@@ -300,8 +301,12 @@ def cmd_jira(ctx: CLIContext, issue_config: str) -> None:
 
 
 @main.command(name='schedule')
+@click.option(
+    '--arch',
+    default='',
```
happz, line 306: `None` would be better, making `arch` optional - it's easier to detect for linters than an empty string. One can easily pass an empty string where non-empty string is expected, but linters will stop us from passing `None` to functions expecting a string.

newa/__init__.py
kkaarreell: Done.

newa/__init__.py
kkaarreell: Fixed.

newa/cli.py
kkaarreell: Fixed.

newa/__init__.py
happz: Both are fine, I for one would put it under the `Arch` class as you did, as a "constructor", under the assumption that the class knows best how to construct its instances from various inputs.

happz (approved):

PR #42, opened 2024-06-18
Default --state-dir uses /var/tmp/newa/run-[0-9]+ pattern

kkaarreell:


newa/cli.py
```diff
@@ -44,10 +45,28 @@
...
 
+def default_state_dir() -> Path:
+    """ Returns the first unused directory matching /var/tmp/newa/run-[0-9]+ """
+    parent_dir = Path('/var/tmp/newa')
+    pattern = '^run-([0-9]+)$'
+    counter = 0
+    try:
+        obj = os.scandir(parent_dir)
+    except FileNotFoundError:
+        return Path(os.path.join(parent_dir, f'run-{counter}'))
```
happz, line 56: `return parent_dir / f'run-{counter}'` should produce the same result, and thanks to `pathlib` magic, it's simpler :)

happz (approved):

PR #48, opened 2024-06-19
Fix Testing Farm integration

kkaarreell:
This is a collection of changes I had to make in order to make the execution in Testing Farm truly work for a container advisory. It is not always split into multiple commits as I was not always able to identify all the changes need to fix something prior fixing something else.

The list of changes includes:
- Before exporting a `Recipe` class to YAML for scheduling render multiple values as Jinja templates
- Introduce `testingfarm` attribute to the `Recipe` class.
- Fix `Arch` object conversion from YAML
- Add `cli_args` handling
- Extend `build_requests()` function with `jinja_vars` argument optionally storing additional variables used for Jinja template rendering (might be necessary for `when` evaluation). 
- Add merging of dimensions having type `str`
- Improve `ExecuteJob` logging
- Handle missing RP `launch_description`
- Added SRPMS to Arch to simplify advisory processing
- Fixed passing undefined variables to Jinja
- Enable debug logging
- Fix unreliable Jira search which may return extra issues

### DIFF
--- a/newa/cli.py
+++ b/newa/cli.py
@@ -508,7 +508,8 @@ def cmd_report(ctx: CLIContext, rp_project: str, rp_url: str) -> None:
             jira_request_mapping[jira_id] = {}
             jira_launch_mapping[jira_id] = RawRecipeReportPortalConfigDimension(
                 launch_name=execute_job.request.reportportal['launch_name'],
-                launch_description=execute_job.request.reportportal['launch_description'])
+                launch_description=execute_job.request.reportportal.get(
+                    'launch_description', None))
             # jira_launch_mapping[jira_id] = execute_job.request.reportportal['launch_name']
         # for each Jira and request ID we build a list of RP launches
         jira_request_mapping[jira_id][request_id] = rp.find_launches_by_attr(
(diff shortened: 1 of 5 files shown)


newa/cli.py
```diff
@@ -360,21 +367,26 @@ def cmd_schedule(ctx: CLIContext, arch: str) -> None:
...
-                        str(request.reportportal.get(rp_attr)),
-                        ERRATUM=jira_job.erratum,
-                        COMPOSE=jira_job.compose,
-                        CONTEXT=request.context,
-                        ENVIRONMENT=request.environment,
-                        )
+            # before yaml export render all fields as Jinja templates
+            for attr in ("reportportal", "tmt", "testingfarm", "environment", "context"):
+                if getattr(request, attr, None):
+                    for (key, value) in getattr(request, attr).items():
```
happz, line 382: You may skip the `getattr` test if you use `{}` as default value, and `getattr()` in assignment might be also good to replace:

```python
for attr in ("reportportal", "tmt", "testingfarm", "environment", "context"):
    mapping = getattr(request, attr, {})

    for (key, value) in mapping.items():
        mapping[key] = render_template(...)
```

If `attr` is not set at all, empty mapping => no iterations, no assignments, no calls to `render_template`, less indentation.

PR #49, opened 2024-06-21
Modify closed issue handling

kkaarreell:
Modify how closed issues are handled. Previously, only opened issues were taken into account and therefore we always ended up with an open issue. That didn't work well for tasks that have been previously "completed" by a user as they were restarted.
This is now changed in the following way:
 - Closed issues are also taken into account and new issue is not created if  matching but closed issue exists
 - Creation of new issues can be enabled through `jira --recreate` option
 - An issue is considered "old" when its parent issue has been re-created and therefore even the issue would be re-created. Previously, this worked only for subtasks but now it works also for other issue types.

### DIFF
--- a/newa/__init__.py
+++ b/newa/__init__.py
@@ -587,6 +587,7 @@ class Issue(Cloneable, Serializable):
     # usually JiraHandler.group takes priority but this value
     # will be used when JiraHandler is not available
     group: Optional[str] = None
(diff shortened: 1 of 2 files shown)


The-Mule (approved):
I did not spot any issue, I did not try it in "production" yet (but I will) but no need to wait for that.

PR #57, opened 2024-06-28
Catch RequestException so we can retry

kkaarreell:


The-Mule (approved):
Reviewed, this will certainly improves robustness. Thanks!

PR #61, opened 2024-07-03
Store TF command in Execution and the YAML file

kkaarreell:


newa/cli.py
```diff
@@ -444,6 +444,11 @@ def worker(ctx: CLIContext, schedule_file: Path) -> None:
     tf_request = schedule_job.request.initiate_tf_request(ctx)
     log(f'TF request filed with uuid {tf_request.uuid}')
 
+    # generate Tf command so we can log it
+    command_args, environment = schedule_job.request.generate_tf_exec_command(ctx)
+    command = ' '.join(command_args)
+    # hide tokens
+    command = command.replace(ctx.settings.rp_token, '***')
```
The-Mule, line 451: It is perhaps out of scope of this PR but we should probably have a secrets section in the settings so that we can filter them all out when needed.

The-Mule (approved):
LGTM

newa/cli.py
kkaarreell: Good point, would you open an issue? And do you have specificidea how it would look like?

PR #66, opened 2024-07-31
Add Erratum.url attribute

kkaarreell:


newa/__init__.py
```diff
@@ -573,6 +574,7 @@ class Erratum(Cloneable, Serializable):  # type: ignore[no-untyped-def]
     summary: str = field(repr=False)
     people_assigned_to: str = field(repr=False)
     release: str = field()
+    url: str = field()
```
kkaarreell, line 577: Since the `url` depends on the value from a different object (ErrataTool) I have decided to have it specified explicitly and do not try to derive the value.
However, we could possible make it an optional argument, although a mandatory one seems to be more reasonable to me.

happz (approved):

PR #69, opened 2024-08-01
Set NO_COLOR and NO_TTY for TF cli

kkaarreell:


matusmarhefka (approved):
Tested and it resolved issues when running `newa` in Gitlab CI pipelines.

kkaarreell:
> Tested and it resolved issues when running `newa` in Gitlab CI pipelines.

Thank you. Merging so that we won't block users.

PR #72, opened 2024-08-07
Enable specifying --context and --environment from CLI

kkaarreell:


newa/__init__.py
```diff
@@ -1380,6 +1380,8 @@ class CLIContext:
     settings: Settings
     # Path to directory with state files
     state_dirpath: Path
+    cli_environment: RecipeEnvironment = field()
+    cli_context: RecipeContext = field()
```
happz, line 1384: `factory=dict` would provide a nice default value

newa/cli.py
```diff
@@ -93,6 +111,27 @@ def main(click_context: click.Context, state_dir: str, conf_file: str, debug: bo
...
+        (k, v) = map(str.strip, s.split('=', 1))
+        # cut-off surrounding " or '
+        if (v.startswith('"') and v.endswith('"')) or (v.startswith("'") and v.endswith("'")):
+            v = v[1:-1]
+        return (k, v)
+
+    # store environment variables and context provided on a cmdline
+    for s in envvars:
+        k, v = _split(s)
+        ctx.cli_environment[k] = v
```
happz, line 129: Maybe a bit a streamlining? `dict()` accepts a list of key/value tuples, `dict([(key1, value1), (key2, value2), ...])`, exactly what `_split()` returns, therefore something like the following would pack it into one line:

```python
ctx.cli_environment.update(dict(_split(s) for s in envvars))
```

newa/__init__.py

newa/cli.py

newa/cli.py
```diff
@@ -93,6 +111,27 @@ def main(click_context: click.Context, state_dir: str, conf_file: str, debug: bo
...
+        """ split key='some value' into a tuple (key, value) """
+        s = s.strip()
+        if ('=' not in s) or s.startswith('='):
+            raise Exception(
+                f'Option --environment {s} has invalid value, key=value format expected!')
+        (k, v) = map(str.strip, s.split('=', 1))
+        # cut-off surrounding " or '
+        if (v.startswith('"') and v.endswith('"')) or (v.startswith("'") and v.endswith("'")):
+            v = v[1:-1]
+        return (k, v)
```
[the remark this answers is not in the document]
kkaarreell: OK. The above didn't work but I went with regexp that I can understand myself and that might do the job as well, more or less. :-)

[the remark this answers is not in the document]
happz: Sure.

One note WRT patterns in Python: raw strings are recommended, they allow for simpler patterns, with less space for typos. And there's also `'` in the pattern, so I would recommend "multiline" string. In this case, we would end up with this: `r"""^\s*([a-zA-Z0-9_][a-zA-Z0-9_\-]*)=["']?(.*?)["']?\s*$"""` - no double backslashes one can forget or misplace.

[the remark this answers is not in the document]
kkaarreell: Updated.

happz (approved):

PR #76, opened 2024-08-14
Wait for TF request completion in the report step

kkaarreell:
Wait for TF requests to completion when newa is issued only with the report step. This way we can re-run newa after the previous command has been terminated in the execute step for whatever reason.

### DIFF
--- a/newa/cli.py
+++ b/newa/cli.py
@@ -3,7 +3,6 @@ import logging
 import multiprocessing
 import os.path
 import re
-import sys
 import time
 from functools import partial
 from pathlib import Path
@@ -705,7 +704,7 @@ def cmd_report(ctx: CLIContext, rp_project: str, rp_url: str) -> None:
 
     # before reporting check if all TF requests are finished
     # unless we have previously executed 'execute' subcommand
-    if 'execute' not in sys.argv:
+    if 'execute' not in ctx.commands_run:
         watchlist = []
         for execute_job in execute_jobs:
             uuid = getattr(getattr(execute_job, 'execution', None),
--- a/newa/cli.py
+++ b/newa/cli.py
@@ -3,6 +3,7 @@ import logging
 import multiprocessing
 import os.path
 import re
+import sys
 import time
 from functools import partial
 from pathlib import Path
@@ -34,6 +35,7 @@ from . import (
     ReportPortal,
     ScheduleJob,
     Settings,
+    TFRequest,
     eval_test,
     get_url_basename,
     render_template,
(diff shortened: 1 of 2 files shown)


kkaarreell:
Hi @happz , I am sure there is things to improve in this PR and I would appreciate your review.

newa/cli.py
```diff
@@ -656,6 +682,21 @@ def cmd_report(ctx: CLIContext, rp_project: str, rp_url: str) -> None:
     if not jira_token:
         raise Exception('Jira URL is not configured!')
 
+    # before reporting check if all TF requests are finished
+    # unless we have previously executed 'execute' subcommand
+    if 'execute' not in sys.argv:
```
kkaarreell, line 687: Is there a better way to check if specific subcommand has been executed in @click? I could also store some variable to `ctx`. The current check is "short" but I am aware of its limits and I would like to replace it with something better.

newa/cli.py
```diff
@@ -624,6 +626,30 @@ def worker(ctx: CLIContext, schedule_file: Path) -> None:
     ctx.save_execute_job('execute-', execute_job)
 
 
+def wait_for_requests_to_finish(ctx: CLIContext, requests: list[TFRequest]) -> None:
+    delay = int(ctx.settings.tf_recheck_delay)
+    watchlist = requests.copy()
+    while watchlist:
+        next_watchlist = []
```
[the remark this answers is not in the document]
happz: I don't think so. Changing the content of a list while iterating over it is not recommended, you may get lucky & things will work, and next Tuesday it might erase your hard drive... Using a "future list" for collecting items and swapping it for `while` to test it looks very appropriate.

newa/cli.py
happz: I don't have any better idea, and it seems simple enough. We could add some `set()` field to `CLIContext` & make every subcommand add its own name to it, that would be cleaner & more obvious. Click does not seem to have an easy way to reach "all previously executed subcommand".

kkaarreell: Thank you, I have pushed an update with such an implementation.

kkaarreell:
I think I should revisit this approach. Waiting in `report` step assumes that all requests have been scheduled and are either running or finished. However, this may not be the case, due to the limit on the number of workers. 
It seems that better approach would be to introduction `execute --continue` which which won't schedule new requests if they have been already scheduled but schedule new requests if not.

kkaarreell:
> I think I should revisit this approach. Waiting in `report` step assumes that all requests have been scheduled and are either running or finished. However, this may not be the case, due to the limit on the number of workers. It seems that better approach would be to introduction `execute --continue` which which won't schedule new requests if they have been already scheduled but schedule new requests if not.

trying to implement this in https://github.com/RedHatQE/newa/pull/81

PR #77, opened 2024-08-16
Introduce Erratum.blocking_builds attribute

kkaarreell:
Adds and populates Erratum.blocking_builds attribute with builds from blocking advisories.


kkaarreell:
I am using this already and it works well. I incline towards merging it, unless @happz @The-Mule have objections.

Also, IIRC @happz you were suggesting some prettier alternative to `{{ ERRATUM.builds|join(' ') }}` but I can't recall now.

PR #79, opened 2024-08-20
Use UPLOAD_TO_LAUNCH instead or RP launch merge

kkaarreell:


kkaarreell:
This is quite a big change and user testing would be very appropriate.

kkaarreell:
After few more days of usage I have identified and just fixed an issue with duplicate RP launch creation when `--continue` is used. Besides that, it seems to work fine.

PR #81, opened 2024-08-26
Introduce execute --continue option

kkaarreell:
Running `newa --state-dir /var/tmp/newa/run-XXX execute --continue report` should enable user to follow up on the previously terminated newa command.


kkaarreell:
This should obsolete the approach used in https://github.com/RedHatQE/newa/pull/76

PR #84, opened 2024-08-30
Add compose mapping support

kkaarreell:
Changes the way how release to compose mapping is implemented.
Allows user to specify custom mapping on cmdline, however in this case regular expressions are not used.

### DIFF
--- a/newa/cli.py
+++ b/newa/cli.py
@@ -188,6 +188,11 @@ def main(click_context: click.Context,
     '--compose-mapping', 'compose_mapping',
     default=[],
     multiple=True,
+    help=('Custom Erratum release to Testing Farm compose mapping in the form '
+          '"RELEASE=COMPOSE". For example, '
+          '"--compose-mapping RHEL-9.4.0.Z.MAIN+EUS=RHEL-9.4.0-Nightly". '
+          'Can be specified multiple times, the 1st match is used'
+          ),
     )
 @click.pass_obj
 def cmd_event(
@@ -225,14 +230,16 @@ def cmd_event(
             r = re.fullmatch(r'([^\s=]+)=([^=]*)', m)
             if r:
                 pattern, value = r.groups()
+                # for regexp=True apply each matching regexp
                 if regexp and re.search(pattern, new_string):
                     new_string = re.sub(pattern, value, new_string)
                     ctx.logger.debug(
                         f'Found match in {new_string} for mapping {m}, new value {new_string}')
+                # for string matching return the first match
                 if (not regexp) and new_string == pattern:
-                    new_string = value
                     ctx.logger.debug(
                         f'Found match in {new_string} for mapping {m}, new value {new_string}')
+                    return value
             else:
                 raise Exception(f"Mapping {m} does not having expected format 'patten=value'")
         return new_string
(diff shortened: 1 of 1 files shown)


newa/cli.py
```diff
@@ -166,13 +204,20 @@ def cmd_event(ctx: CLIContext, errata_ids: list[str], compose_ids: list[str]) ->
...
-                if compose.endswith('.GA'):
-                    compose = compose[:-3]
-                compose += '-Nightly'
-                # handle compose differences between ET and TF
-                compose = compose.replace('RHEL-10.0.BETA', 'RHEL-10-Beta')
+                release = erratum.release.strip()
+                # when compose_mapping is provided, apply it
+                if compose_mapping:
+                    compose = apply_mapping(release, compose_mapping, regexp=False)
+                else:
```
lukaszachy, line 211: Isn't this branch unnecessary? `apply_mapping` takes care of the empty compose_mapping itself.

newa/cli.py
```diff
@@ -155,6 +165,34 @@ def cmd_event(ctx: CLIContext, errata_ids: list[str], compose_ids: list[str]) ->
     if not errata_ids and not compose_ids:
         raise Exception('Missing event IDs!')
 
+    def apply_mapping(string: str,
+                      mapping: Optional[list[str]] = None,
+                      regexp: bool = True) -> str:
+        # define default mapping
+        if not mapping:
+            mapping = [
+                '\\.GA$=',
```
lukaszachy, line 174: raw strings instead? No need to escape that much

lukaszachy (commented):
I'm missing user facing documentation/examples. IIUIC the mapping is in `<regexp1>=<regexp2>` and mapping is applied for first matching regexp1.

newa/cli.py
kkaarreell: Good point!

newa/cli.py
kkaarreell: The major difference here is different `regexp` value. I could remove the list, true.

newa/cli.py
```diff
@@ -155,6 +165,34 @@ def cmd_event(ctx: CLIContext, errata_ids: list[str], compose_ids: list[str]) ->
...
+                if regexp and re.search(pattern, new_string):
+                    new_string = re.sub(pattern, value, new_string)
+                    ctx.logger.debug(
+                        f'Found match in {new_string} for mapping {m}, new value {new_string}')
+                if (not regexp) and new_string == pattern:
+                    new_string = value
+                    ctx.logger.debug(
+                        f'Found match in {new_string} for mapping {m}, new value {new_string}')
+            else:
+                raise Exception(f"Mapping {m} does not having expected format 'patten=value'")
```
happz, line 193: You can reverse the condition and save one indentation level:

```python
if not r:
  raise Exception(...)
```

newa/cli.py
kkaarreell: Updated

newa/cli.py
kkaarreell: Updated

kkaarreell:
> I'm missing user facing documentation/examples. IIUIC the mapping is in `<regexp1>=<regexp2>` and mapping is applied for first matching regexp1.

I have added more comments and a help string. For user cli usage no regexp is used ATM, I think it would be rather an obstacle. regexp are used only internally for the built-in mapping.

newa/cli.py
kkaarreell: Updated

PR #87, opened 2024-08-31
Add --prev-state-dir option

kkaarreell:
Introduces `--prev-state-dir` option so that user doesn't have to provide `--state-dir` manually and instead the most recent state-dir used within the same shell session is used.

### DIFF
--- a/newa/cli.py
+++ b/newa/cli.py
@@ -1,7 +1,7 @@
 import datetime
 import logging
 import multiprocessing
-import os.path
+import os
 import re
 import time
 from functools import partial
@@ -41,6 +41,8 @@ from . import (
     )
 
 JIRA_NONE_ID = '_NO_ISSUE'
+STATEDIR_PARENT_DIR = Path('/var/tmp/newa')
+STATEDIR_NAME_PATTERN = r'^run-([0-9]+)$'
 
 logging.basicConfig(
     format='%(asctime)s %(message)s',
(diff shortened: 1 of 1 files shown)


newa/cli.py
```diff
@@ -41,35 +41,55 @@
     )
 
 JIRA_NONE_ID = '_NO_ISSUE'
+STATEDIR_PARENT_DIR = '/var/tmp/newa'
```
happz, line 44: It should be set to a `Path` instance right here, to prevent any path-like string sneaking in.

newa/cli.py
```diff
@@ -41,35 +41,55 @@
...
+STATEDIR_NAME_PATTERN = '^run-([0-9]+)$'
```
happz, line 45: I'd recommend making this a raw string. There are no regex features that would be affected, but it's a good habit to build.

newa/cli.py
```diff
@@ -41,35 +41,55 @@
...
             c = int(r.group(1))
-            if c >= counter:
+            if use_ppid:
+                ppid_file = parent_dir / entry.name / ppid_filename
+                if ppid_file.exists() and c >= counter:
+                    counter = c
+            elif c >= counter:
                 counter = c + 1
-    return parent_dir / f'run-{counter}'
+    if counter > 0:
```
happz, line 78: `if counter:` is a more pythonic version, `0` is a false-ish value.

newa/cli.py

newa/cli.py

newa/cli.py

kkaarreell:
I have realized the algorithm is not correct so I have modified it slightly. Also, I think it might be easier to understand now.

PR #92, opened 2024-09-04
Bring back ENVIRONMENT to cmd_jira

ukulekek:
It was originally added as a part of https://github.com/RedHatQE/newa/pull/72, but I believe it was accidentally removed recently.


kkaarreell (approved):
Thank you.

PR #96, opened 2024-09-10
Add list subcommand

kkaarreell:
Adds `list` subcommand for users to list details of the most recent `newa` executions.
Also extends `class Jira()` with new optional attributes so that these attributes can be printed on screen.

### DIFF
--- a/newa/cli.py
+++ b/newa/cli.py
@@ -3,6 +3,7 @@ import logging
 import multiprocessing
 import os
 import re
+import sys
 import time
 from collections.abc import Generator
 from functools import partial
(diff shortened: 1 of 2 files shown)


kkaarreell:
I know, those `print` indentations are ugly, I will wrap it with another function.

newa/cli.py
```diff
@@ -148,6 +149,11 @@ def main(click_context: click.Context,
         )
     click_context.obj = ctx
 
+    # this is here just to suppress state-dir creation
+    # and further processing in case of 'list' subcommand
+    if 'list' in sys.argv:
+        return
+
```
The-Mule, line 156: Can we move it after debug logging setup (I can see a valid use case to call it with --debug)?

The-Mule (approved):

newa/cli.py
kkaarreell: Done. Also preserved set DEBUG level in the `list` subcommand.

happz (approved):

PR #98, opened 2024-09-13
Update Jira issue with RP launch URL

kkaarreell:
Update Jira issue with URL of a newly created (empty) ReportPortal launch so that Jira issue followers know that tests have been scheduled and where to check for updates


kkaarreell:
I am a bit consolidating Jira connection usage in cli.py but it would be nice to rework Jira connection parts even more.
I have filed https://github.com/RedHatQE/newa/issues/99

PR #100, opened 2024-09-16
Fix default architecture handling for compose

kkaarreell:
Previously, `Architectures` was not populated as a list of `Arch` objects when loaded from erratum.
This was not consisted with other assignments. This is now unified.
In addition, change the default `architectures()` return value to a full set of architectures.

### DIFF
--- a/newa/__init__.py
+++ b/newa/__init__.py
@@ -330,7 +330,7 @@ class Arch(Enum):
         _all = [Arch(a) for a in Arch.__members__.values() if a not in _exclude]
 
         if not preset:
-            return [Arch('x86_64')]
+            return _all
         # 'noarch' should be tested on all architectures
         if Arch('noarch') in preset:
             return _all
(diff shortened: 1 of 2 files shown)


newa/cli.py
```diff
@@ -688,8 +688,10 @@ def cmd_schedule(ctx: CLIContext, arch: list[str]) -> None:
             architectures = Arch.architectures(
                 [Arch(a.strip()) for a in arch])
         else:
-            architectures = jira_job.erratum.archs if (
```
kkaarreell, line 691: Hi @happz , here I need your help. According to my testing, when loading `archs` from a YAML file then `jira_job.erratum.archs` is a list of strings while it is supposed to be a list of `Arch`. I believe the root cause is the missing converter for `archs` attribute which I am adding above but please review whether my assumption is correct. Functional-wise it seem to work.

Also, it seems that the same problem is with other attributes like
https://github.com/RedHatQE/newa/blob/main/newa/__init__.py#L798
which would explain why I need to access those as dictionary keys and not object attributes. Am I right?

newa/__init__.py
```diff
@@ -614,7 +614,10 @@ class Erratum(Cloneable, Serializable):  # type: ignore[no-untyped-def]
     people_assigned_to: str = field(repr=False)
     release: str = field()
     url: str = field()
-    archs: list[Arch] = field(factory=list)
+    archs: list[Arch] = field(factory=list,  # type: ignore[var-annotated]
```
kkaarreell, line 617: Hi @happz , here I need your help. According to my testing, when loading archs from a YAML file then jira_job.erratum.archs at
https://github.com/RedHatQE/newa/blob/main/newa/cli.py#L691
I get a list of strings while it is supposed to be a list of Arch. I believe the root cause is the missing converter for archs attribute which I am adding above but please review whether my assumption is correct. Functional-wise it seem to work.

happz: I believe you are correct, there are several other instances of this approach.

PR #108, opened 2024-10-10
ReportPortal: Jira ID text to hyperlink

### LINKED ISSUE #107: ReportPortal: Jira ID text to hyperlink
RP portal launch contains text with ID of Jira main epic issue.
It is not a link to that Jira system, it is just a text.

It would be good to have hyperlink.

`__cli__.py` in TF REQ rection:

> it would be nice to use hyperlinks in launch description however we
> would hit description length limit. Therefore using plain text

I do not think this apply for Jira.

jharuda:
Resolves #107 
Screenshot from 2024-10-11 00-53-30

### DIFF
--- a/newa/cli.py
+++ b/newa/cli.py
@@ -5,6 +5,7 @@ import os
 import re
 import sys
 import time
+import urllib
 from collections.abc import Generator
 from functools import partial
 from pathlib import Path
@@ -822,6 +823,7 @@ def cmd_execute(ctx: CLIContext, workers: int, _continue: bool) -> None:
     # store all launch uuids for later finishing
     launch_list = []
     # now we process jobs for each jira_id
+    jira_url = ctx.settings.jira_url
     for jira_id in jira_schedule_job_mapping:
         # when --continue the launch was probably already created
         # check the 1st job for launch_uuid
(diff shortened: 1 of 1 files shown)


kkaarreell:
Thank you for the change. My only concern is that RP has quite small limit (512 chars I think) on launch description and with hyperlink usage we can hit this limit more easily. Therefore I have decided not to use hyperlinks for TF requests. OTOH, there would be only one Jira issue so I think it is safe to do the change.

newa/cli.py
```diff
@@ -844,7 +845,7 @@ def cmd_execute(ctx: CLIContext, workers: int, _continue: bool) -> None:
             launch_description += '<br><br>'
         # add the number of jobs
         if not jira_id.startswith(JIRA_NONE_ID):
-            launch_description += f'{jira_id}: '
+            launch_description += f'[{jira_id}]({jira_url}/browse/{jira_id}): '
```
kkaarreell, line 848: Could you maybe import `urllib` and use
```
issue_url = rllib.parse.urljoin(
                jira_url,
                f"/browse/{jira_id}")
launch_description += f'[{jira_id}]({issue_url}): '
```
? In the code we tried to avoid building URLs like this.

jharuda: OK, I updated it and used `urllib`.

newa/cli.py
```diff
@@ -41,7 +42,7 @@
     eval_test,
     get_url_basename,
     render_template,
-    )
+)
```
jharuda, line 45: Github action/lint recommended this.

kkaarreell (approved):

PR #116, opened 2024-11-01
Use global TF request counter in request ID

kkaarreell:


mildas (approved):
Tested with COPR build and works fine.

PR #120, opened 2024-11-07
Render recipe url

kkaarreell:


lukaszachy (approved):
Thanks, for as I expect

```
# jira config contains
job_recipe: "{{ ENVIRONMENT.job_recipe_url }}"

$ newa -e job_recipe_url=https://adhoc ... jira ...
# and jira*yaml then contains
url: 'https://adhoc'
```

PR #122, opened 2024-11-07
Add support for issue state transitions

kkaarreell:
Adds configuration for two new Jira issue transition. `processed` is used when issue action processing is completed (including optional test execution and reporting). `passed` is used when tests have been executed and all passed.
State transition is enabled in the issue action through `auto_transition: True`.


lukaszachy (approved):
Thanks for implementing this so fast.

PR #124, opened 2024-11-14
Add basic support for issue-config file defaults

kkaarreell:
Adds support for issue config file default settings. Such a settings is applied to individual actions defined under `issues` attribute.
Example:
```
defaults:
  assignee: null
  fields:
    "Pool Team": "my_great_team"
    "Story Points": 0
```
`defaults` are merged when defined in multiple (included) files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,81 @@ TESTING_FARM_API_TOKEN
 NEWA_TF_RECHECK_DELAY
 ```
 
+## Jira issue configuration file
+
+This is a configuration for the `newa jira` subcommand and typically it targets a particular package and event, e.g. it prescribes which Jira issues should be created for an advisory.
+This configuration file is passed to newa command like this:
+```
+newa ... jira --issue-config CONFIG_FILE ...
+```
+Issue config file may utilize Jinja2 templates in order to adjust configuration with event specific data.
+
+Example of the Jira issue configuration file:
+```
+include: global_settings.yaml
+
+project: MYPROJECT
+
+transitions:
+  closed:
+    - Closed
+  dropped:
+    - Closed.Obsolete
+  processed:
+    - In Progress
+  passed:
+    - Closed.Done
+
+defaults:
+   assignee: '{{ ERRATUM.people_assigned_to }}'
+#  fields:
+#    "Pool Team": "my_great_team"
+#    "Story Points": 0
+
+issues:
+
+ - summary: "Errata Workflow Checklist {% if ERRATUM.respin_count > 0 %}(respin {{ ERRATUM.respin_count }}){% endif %}"
+   description: "Task tracking particular respin of errata."
+   type: task
+   id: errata_task
+   parent_id: errata_epic
+   on_respin: close
+```
+
+Individual settings are described below.
+
+### include
+
+Allows user to import snippet of a file from a different URL location or a file.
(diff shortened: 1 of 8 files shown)


newa/cli.py
```diff
@@ -511,13 +521,35 @@ def _jira_fake_id_generator() -> Generator[str, int, None]:
...
+
             # Iterate over issue actions. Take one, if it's not possible to finish it,
             # put it back at the end of the queue.
             while issue_actions:
                 action = issue_actions.pop(0)
 
                 ctx.logger.info(f"Processing {action.id}")
 
+                # update action object with default attributes when not present
+                action = _update_action_with_defaults(action, config.defaults)
```
lukaszachy, line 551: Why is handling of defaults done here in the loop and not during `config = IssueConfig.from_yaml_with_include` call?

kkaarreell: Well, I guess I see that function for loading YAML, not doing semantic evaluation. Are you strongly against keeping it here?

lukaszachy: > Well, I guess I see that function for loading YAML, not doing semantic evaluation.

That function already extends YAML loader with 'include'  and isn't the 'default' closer to syntax? From YAML perspective it is just a default dictionary to update with whatever it reads from the rest of the file.

IMO 'issue_actions' should contain "final" classes before they reach 'while' loop.

kkaarreell: I have pushed an update, pls review.

lukaszachy (approved):
Works great, however more documentation will be appreciated - seems that if multiple layers of 'include' are used, the `defaults` is used from the top one (closest to the file used in `read_file()` call)

newa/cli.py
lukaszachy: I don't see 'resolve' button so please mark this thread as resolved.

kkaarreell:
> Works great, however more documentation will be appreciated - seems that if multiple layers of 'include' are used, the `defaults` is used from the top one (closest to the file used in `read_file()` call)

I see. I think that might be relevant for `include` processing in general. Ideally, users would avoid conflicts. I will test the behavior a bit more and either document it or modify it (if reasonably doable).

kkaarreell:
> > Works great, however more documentation will be appreciated - seems that if multiple layers of 'include' are used, the `defaults` is used from the top one (closest to the file used in `read_file()` call)
> 
> I see. I think that might be relevant for `include` processing in general. Ideally, users would avoid conflicts. I will test the behavior a bit more and either document it or modify it (if reasonably doable).

I have pushed and update changing how the merging is done. Also, I have updated README with more notes on the topic.
I do also plan to add a test testing it.

README.md
```diff
@@ -46,6 +46,74 @@ TESTING_FARM_API_TOKEN
...
+The only exceptions are are `issues` and `defaults` which are merged.
+To unset a value defined in an included file one can set the value to `null`.
+
+### project
+
+Defines Jira project to be used by NEWA.
+
+### transitions
+
+This is a mapping which tells NEWA which particular issue states (and resolutions) it should be using. This settings depends on a particular Jira project. It is also possible to specify resolution separated by a dot, e.g. `Closed.Obsolete`.
```
lukaszachy, line 106: Could you enumerate which transitions are required and optional?

lukaszachy (approved):

PR #132, opened 2024-11-29
Adjust quoting due to TF CLI change

kkaarreell:
See
https://docs.testing-farm.io/Testing%20Farm/0.1/cli.html#_breaking_change_for_strings_with_whitespaces


kkaarreell:
tested, works fine

PR #138, opened 2024-12-07
Add --extract-state-dir option

kkaarreell:
The purpose of this option is to enable easy rescheduling using data from a previous newa job executed by an automation (e.g. Jenkins).

### DIFF
--- a/newa/cli.py
+++ b/newa/cli.py
@@ -1,9 +1,11 @@
 import datetime
+import io
 import logging
 import multiprocessing
 import os
 import re
 import sys
+import tarfile
 import time
 import urllib
 from collections.abc import Generator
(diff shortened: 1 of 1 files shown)


newa/cli.py
```diff
@@ -189,6 +197,19 @@ def main(click_context: click.Context,
     if not ctx.state_dirpath.exists():
         ctx.logger.debug(f'State directory {ctx.state_dirpath} does not exist, creating...')
         ctx.state_dirpath.mkdir(parents=True)
+
+    # extract YAML files from the given archive to state-dir
+    if extract_state_dir:
+        if re.match('^http[s]?:', extract_state_dir):
```
happz, line 203: Out of curiosity, why `[s]?`, wouldn’t `s?` work as well?

kkaarreell: Just my habit. Fixed. Thank you for a review.

Issue #148, opened 2024-12-16
`newa list` traceback with custom state-dir name


lukaszachy:
I created by own state-dir name (`newa --state-dir=/var/tmp/newa/12345`) and `newa list` now fails with

```
  File "/usr/lib/python3.13/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/usr/lib/python3.13/site-packages/click/decorators.py", line 45, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/usr/lib/python3.13/site-packages/newa/cli.py", line 254, in cmd_list
    sorted_entries = sorted(entries, key=lambda entry: int(entry.name.split('-')[1]))
  File "/usr/lib/python3.13/site-packages/newa/cli.py", line 254, in <lambda>
    sorted_entries = sorted(entries, key=lambda entry: int(entry.name.split('-')[1]))
```

I obviously broke the expected name pattern (foo-number).
How about making newa work with custom names? Or failing when creating them?

kkaarreell:
It is possible. Just `--prev-state-dir` option won't work properly.

kkaarreell:
A traceback should be fixed with https://github.com/RedHatQE/newa/pull/152.
That custom state-dir path would be ignored by `newa list` though.
We would need something like `.bash_history` to make it work with custom state-dir locations properly.

kkaarreell:
@lukaszachy Could you please newa version from PR#152? Thank you.

The-Mule:
(Lukas, ignore comment above about custom state dirs being ignored, even though it was true with the original version #152 it is no longer true)

lukaszachy:
@kkaarreell no traceback raised for state-dir called `/var/tmp/newa/12345`, looks fixed to me:)

PR #152, opened 2025-01-02
Let 'newa list' sort dirs with `getctime()`

kkaarreell:


The-Mule:
If I understand correctly the only problem causing #148 is the sorting logic. Can't we sort directories by the time of their creation instead of their names? That would allow using custom state directories and keep all commands working (as long as directories are placed in /var/tmp/newa), right? 

With this update the idea of using custom state directories starts going odd direction since you can have it but e.g. `list` won't be using them and from my experience `list` command is pretty useful and hard to live without...

The-Mule (approved):
Right, I suppose this is a more universal approach. Thanks!

kkaarreell:
> If I understand correctly the only problem causing #148 is the sorting logic. Can't we sort directories by the time of their creation instead of their names? That would allow using custom state directories and keep all commands working (as long as directories are placed in /var/tmp/newa), right?
> 
> With this update the idea of using custom state directories starts going odd direction since you can have it but e.g. `list` won't be using them and from my experience `list` command is pretty useful and hard to live without...

Originally, I was thinking about the modification time, but that could provide weird outputs in case there would be multiple newa commands writing updates in parallel. Creation time looks better. However, `get_state_dir()` needs to be modified as well.

Issue #151, opened 2024-12-20
Support for re-run of REQs


The-Mule:
Right now there is an option to rerun REQs based on their result (e.g. all that failed). It would be very practical to be able to specify exactly which REQs to run, e.g.:

```
newa execute --restart-reqs  REQ-1.2.3, REQ-4.5.6
```

kkaarreell:
This is being worked on in https://github.com/RedHatQE/newa/pull/154.
ATM, one cannot use the short for `--restart-req REQ-1.2.1,REQ-2.2.2`, however you could use regexp.
Providing multiple values with just one switch is not yet currently supported. We could use that also for other `newa` options but better to do it in a separate request (or commit).

kkaarreell:
@The-Mule could you please test it?

The-Mule:
Tested both long and short variants using it both once and twice (to restart two reqs at the same time), it works. The very last version dropped support of regular expression but I think it is not really needed since you can use -R multiple times and the structure of REQ ids does not really make much difference from that and RE. Thank you for this functionality. 

It is super useful, especially when paired with recipe having a single dimension matrix for each ("leaf") plan (resulting of one REQ per plan and architecture).

PR #154, opened 2025-01-02
Add execute --restart-request option

kkaarreell:


kkaarreell:
I am not sure about the `/regexp/` variant. I think it won't hurt if the passed string (e.g. `REQ-1.2.1`) would be always processed as regexp . Despite `.` can match any symbol, due to having digits in place and `re.fullmatch()` usage, in reality it should not match anything else... unless u user does a typo. What I like about using `/` is that it's explicit, in NEWA it is a new concept though.

The-Mule:
Yeah, I am also not sure about RE. On one hand I like it, on the other hand I feel like it is not necessary. To me the idea of REQ re-execution is to be able to re-execute isolated REQs and since you can already use the option multiple times I don't really see much need for allowing RE (also, it feels like it would be easier to list the REQs explicitly rather than coming up with the right RE to match most of them). I appreciate the idea of distinguishing RE by `/.../` but it needs a proper help description that is currently missing. All in all, It is a nice touch but I don't think it will be used much, for example this was what I had to do in my case:

```
newa --state-dir /var/tmp/newa/run-10 execute --restart-request '/REQ-4[38].124.4[38]/'
```

while the following variant is more readable and requires no RE implementation in this PR:

```
newa --state-dir /var/tmp/newa/run-10 execute --restart-request REQ-43.124.43 --restart-request REQ-48.124.48
```

Following the KISS principle I would slightly prefer "dumb" version without RE but I don't really mind it. I briefly tested it and it works fine. I think this is a really important addition that will be used a lot!

kkaarreell:
OK, I have removed the regexp support. Btw, note the recently added short version `-R`.

kkaarreell:
@The-Mule Would you mind testing it one more time? Thank you.

kkaarreell:
> @The-Mule Would you mind reviewing&testing it one more time? Thank you.

testing feedback provided in https://github.com/RedHatQE/newa/issues/151#issuecomment-2569002744

PR #158, opened 2025-01-14
Update README with event and jira options.

kkaarreell:


kkaarreell:
Hi @happz , what would be a good way of fixing the remaining pre-commit issue? Do I have to create a function to avoid duplicating a code in `with ..` expressions?

happz:
> Hi @happz , what would be a good way of fixing the remaining pre-commit issue? Do I have to create a function to avoid duplicating a code in `with ..` expressions?

Either that, or you can collect kwargs:

```python
        tar_open_kwargs: dict[str, Any] = {
            'mode': 'r:*',
        }

        if re.match('^https?://', extract_state_dir):
            data = urllib.request.urlopen(extract_state_dir).read()
            tar_open_kwargs['fileobj'] = io.BytesIO(data)
        else:
            tar_open_kwargs['name'] = Path(extract_state_dir)

        with tarfile.open(**tar_open_kwargs) as tf:
            for item in tf.getmembers():
                if item.name.endswith('.yaml'):
                    item.name = os.path.basename(item.name)
                    tf.extract(item, path=ctx.state_dirpath, filter='data')
```

README.md
```diff
@@ -645,6 +645,15 @@ request:
   tmt_path: ''
 ```
 
+#### Option `--arch`
+
+By default, tests are scheduled for all relevant architectures. This option can be used to limit scheduling to a particula architecture. This option can be used multiple times.
```
The-Mule, line 650: Typo: particula -> particular

kkaarreell: fixed

PR #164, opened 2025-02-03
NEWA_COMMENT_FOOTER might not be set

lukaszachy:


lukaszachy:
hm, no idea why mypy complains... `newa/cli.py:1158: error: "_Environ[str]" not callable  [operator]`

kkaarreell (approved):

PR #167, opened 2025-02-06
Do TF request status update before reporting

kkaarreell:


jharuda:
I tested it and it works fine.

LGMT.

PR #173, opened 2025-03-03
Update execute_job with artifacts_url when known

kkaarreell:
When TF is initiated, `artifacts_url` is not yet known. This change updates `execute_job` with `artifacts_url` when it becomes known so that commands like `newa list` can present it to a user.

### DIFF
--- a/newa/cli.py
+++ b/newa/cli.py
@@ -1302,6 +1302,13 @@ def worker(ctx: CLIContext, schedule_file: Path) -> None:
         tf_request.fetch_details()
         if tf_request.details:
             state = tf_request.details['state']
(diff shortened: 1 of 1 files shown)


The-Mule:
Thanks a lot, works fine for me!

The-Mule (approved):

The-Mule:
Just to double check - when running with --no-wait, I need to do 'execute --continue --no-wait' before 'list' to refresh the results, right?

kkaarreell:
> Just to double check - when running with --no-wait, I need to do 'execute --continue --no-wait' before 'list' to refresh the results, right?

Yes. This is because artifacts_url is not known since the very beginning and it is available through API only once a request is being processed by TF.

Issue #179, opened 2025-03-18
Support 'Sprint' in jira


lukaszachy:
API requires sprint id, which is difficult to use. Idea for now is to allow 'placeholder' value eg 'active' and detect the active sprint from board id.

Bits & pieces:

```
# Get active sprint from board name
def get_active_sprint_id(self, board_name: str ) -> int:
    boards = jira.boards(name=board_name)
    assert len(boards) == 1
    board_id: int = boards[0].id
    # Limit to sprints from requested board
    active_sprints = [
        s for s in jira.sprints(board_id, state='active') if s.originBoardId == board_id
    ]
    return active_sprints[0].id
```
"board_name" could be part of jira newa config file, `sprint: active` or `sprint: 12345` to be as part of issue-config or command line.

PR #180, opened 2025-03-18
Allow to assign Sprint using fields.Sprint

lukaszachy:
For now just using fields. 

I've filled #179 for better UX, but lets have something working right now.


lukaszachy:
Dropping in favour of #183

PR #191, opened 2025-03-25
Add --prev-event and --prev-issue newa options

kkaarreell:


jharuda:
I tested it and it works fine.

LGTM.

PR #212, opened 2025-04-30
Do not exit after 'list' subcommand

kkaarreell:
In fact it may be handy to be able to run `list` subcommand after `execute` or `report` to get more pretty status report.

Example:
newa ... execute report list
or
newa ... execute list report

Also, previously it could happen that some newa commands created a new state-dir even though it was not needed. This should not happen now, commands like `list` `cancel` and options `--help` are not creating new state-dir.


Koncpa (approved):

Koncpa:
RPMbuild manually tested, code LGTM!

PR #219, opened 2025-05-12
Expose 'revision' field of Erratum

lukaszachy:
If builds change in NEW_FILES the respin_count isn't incremented

Errata didn't leave NEW_FILES, builds where changed:

```
  "revision": 2,
  "status": "NEW_FILES",
  "respin_count": 0,
```

## Summary by Sourcery

Enhancements:
- Expose the 'revision' field on Erratum by parsing it from info_json and including it in the data class

### DIFF
--- a/newa/__init__.py
+++ b/newa/__init__.py
@@ -689,6 +689,8 @@ class ErrataTool:
                         content_type=content_type,
                         respin_count=int(
                             info_json["respin_count"]),
+                        revision=int(
+                            info_json["revision"]),
                         summary=info_json["synopsis"],
                         release=release,
                         builds=builds,
@@ -779,6 +781,7 @@ class Erratum(Cloneable, Serializable):  # type: ignore[no-untyped-def]
     people_package_owner: Optional[str] = None
     people_qe_group: Optional[str] = None
     people_devel_group: Optional[str] = None
+    revision: Optional[int] = field(repr=False, default=0)
 
 
 @define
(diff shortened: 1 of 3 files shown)


lukaszachy:
> 🟢 Testing: all looks good

If only CI didn't fail with 'missing required argument' because I forgot to update tests/unit/test_eval_test.py ...

kkaarreell:
@lukaszachy Thank you. Could you please try to run `newa` command with yaml data made by the older version? Something skipping the `event` subcommand like `newa -D some/old/run jira ...`. I think that it could traceback because of missing `revision`. Probably it should be made optional and assigned a default value (None). We do not want `newa` to crash on old data.

kkaarreell:
> > 🟢 Testing: all looks good
> 
> If only CI didn't fail with 'missing required argument' because I forgot to update tests/unit/test_eval_test.py ...

Yeah, this is what I mean in my other comment. We need to fix `newa`, not a test. :-D

kkaarreell (changes requested):
Pls make the attribute optional so `newa` won't crash when it's missing in YAML.

lukaszachy:
> Pls make the attribute optional so `newa` won't crash when it's missing in YAML.

Done in bd22b97 but I have no idea how to fix the other way (it is there but older newa doesn't want it).
`TypeError: Erratum.__init__() got an unexpected keyword argument 'revision'`

I set to '0' if missing to avoid the need to catch for 'None' during prints.

kkaarreell (approved):

PR #369, opened 2026-06-01
Add architecture info to list command TF request output

### LINKED ISSUE #368: Add arch info into "newa list" output
If we're listing the requests from state-dir and respective data like state via `newa -D <state-dir> list`, result and artifacts URL, it looks like this

`REQ-X.Y.Z - state: <state>, result: <result>, artifacts: <URL>`

Once we're taking care of multiple requests during f.e CTC or other types of testing, it would be handy to also display the respective **architecture** into the output of the list of requests. It's a minor change but you are directly able to navigate into a specific artifacts for that architecture if you recall which requests has which architecture.

Output could look like this

`REQ-X.Y.Z - state: <state>, result: <result>, architecture: <architecture>, artifacts: <URL>`

kkaarreell:
Display architecture alongside artifacts URL in 'newa list' output to help users quickly identify which architecture a test request was executed on.

Resolves: https://github.com/RedHatQE/newa/issues/368

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

New Features:
- Display the architecture of each test request in the `newa list` command output when available.

### DIFF
--- a/newa/cli/commands/list_cmd.py
+++ b/newa/cli/commands/list_cmd.py
@@ -221,13 +221,16 @@ def cmd_list(
                                     execute_job.execution, "result", RequestResult.NONE)
                                 url = getattr(
                                     execute_job.execution, "artifacts_url", "not available")
+                                # Get architecture if available
+                                arch = getattr(execute_job.request, "arch", None)
+                                arch_suffix = f', arch: {arch.value}' if arch else ''
                                 # Apply color formatting to state and result
                                 colored_state = colorize_state(state)
                                 colored_result = colorize_result(result.value)
                                 print(
                                     f' - state: {colored_state}, '
                                     f'result: {colored_result}, '
-                                    f'artifacts: {url}')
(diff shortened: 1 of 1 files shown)


kkaarreell:
> Hey - I've left some high level feedback:
> 
>     * When building `arch_suffix`, you assume `arch` has a `.value` attribute; if `arch` can be a plain string or another type in some cases, consider using `str(arch)` (or handling both Enum and non-Enum) to avoid potential attribute errors.
> 
> 
> Prompt for AI Agents
> 
> Sourcery is free for open source - if you like our reviews please consider sharing them ✨
> 
> 
> Help me be more useful! Please click 👍 or 👎 on each comment and I'll use the feedback to improve your reviews.

  The converter=Arch ensures that any value assigned to arch is automatically converted to an Arch enum. This means arch should only ever be:
  1. An Arch enum instance, or
  2. None

PR #371, opened 2026-06-10
Add on_respin='inherit' feature for IssueAction

kkaarreell:
Implement new 'inherit' value for the on_respin attribute that allows child issues to copy the on_respin behavior from their parent issue.

Changes:
- Add INHERIT enum value to OnRespinAction
- Implement resolution logic in _process_issue_action() that:
  * Validates parent_id exists when using 'inherit'
  * Copies parent's on_respin value to child during processing
  * Includes defensive validation and debug logging
- Update README with documentation and example
- Add comprehensive unit tests covering inheritance scenarios and error cases

The resolution happens during action processing (not during config loading), ensuring parents are processed before children via the existing queueing mechanism. Chain inheritance is supported (grandchild -> child -> parent).

## Summary by Sourcery

Introduce inheritance-based on_respin behavior for child issues in Jira processing.

New Features:
- Allow issues to specify on_respin='inherit' to copy respin behavior from their parent issue.

Enhancements:
- Document the new on_respin inheritance behavior and defaults in the README.

Tests:
- Add unit tests and sample configs validating on_respin inheritance resolution and error handling for missing or invalid parents.


kkaarreell:
>     * The parent lookup in `_process_issue_action` linearly scans `config.issues` for each child; consider using a precomputed `{id: action}` mapping or a helper to avoid repeated O(n) searches as configurations grow.

The number of ActionItems objects is rather small so there is no big impact on performance.

>     * The defensive check that raises when both child and parent have `on_respin='inherit'` conflicts with the documented support for chain inheritance and with the recursive resolution used in the tests; it would be more robust to resolve the parent’s inherit value here instead of raising.

We're working on the same config.issues list throughout processing. When we process the child and mutate action.on_respin = parent.on_respin,
  we're mutating the actual IssueAction object in config.issues. So when the grandchild later looks up its parent (the child), it sees the already-resolved
  value, not the original inherit.

PR #380, opened 2026-06-13
Fix i386 architecture excluded from RHEL-6 test scheduling

brozs:
## Summary

- Refactor `Arch.architectures()` to use explicit whitelists (`_valid`, `_default`) instead of an exclude list, and add a RHEL-6 compose check that includes i386 as a valid test architecture
- Pass `compose` to `Arch.architectures()` in `errata_service.py` (via `derive_compose()`) and `schedule_helpers.py` so the RHEL-6 logic activates
- Add unit tests for RHEL-6 i386 inclusion, noarch/multi fallback, and confirming i386 stays excluded for non-RHEL-6

## Problem

When newa processes an errata advisory for RHEL-6 (e.g. ER#168066, `RHEL-6-ELS-EXTENSION`), the i386 architecture is dropped from the arch list even though the ET product listings include i686 RPMs. The ET API returns `i386` as the channel key, but `Arch.architectures()` unconditionally filtered it out. No tests were scheduled for i386 on RHEL-6, where it is a first-class architecture with standalone composes.

## Test plan

- [x] All 271 existing unit tests pass
- [x] 5 new tests in `test_arch.py` cover RHEL-6 defaults, preset intersection, noarch/multi fallback, and non-RHEL-6 exclusion
- [x] Ruff and mypy clean (no new warnings)
- [x] Verified with COPR build (`sbroz/newa-testing`) against advisory 168066 — i386 now appears in scheduled arches

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Allow i386 to be treated as a valid test architecture for RHEL-6 composes while keeping it excluded elsewhere, and ensure compose-aware architecture selection is used when scheduling tests and fetching errata.

New Features:
- Enable i386 as a valid and default test architecture specifically for RHEL-6 composes.

Bug Fixes:
- Fix omission of i386 tests for RHEL-6 advisories where i386 is a first-class architecture.

Enhancements:
- Refine architecture selection to use explicit valid/default architecture whitelists with compose-specific behavior.
- Propagate compose information into architecture resolution in errata processing and scheduling helpers.

Tests:
- Add unit tests covering RHEL-6 default architectures, preset inclusion of i386, noarch/multi fallbacks, and i386 exclusion for non-RHEL releases.

### DIFF
--- a/newa/cli/schedule_helpers.py
+++ b/newa/cli/schedule_helpers.py
@@ -29,7 +29,7 @@ def _determine_architectures(
         compose: Optional[str]) -> list[Arch]:
     """Determine which architectures to use for scheduling."""
     if arch_options:
-        return Arch.architectures([Arch(a.strip()) for a in arch_options])
+        return Arch.architectures([Arch(a.strip()) for a in arch_options], compose=compose)
 
     if jira_job.erratum and jira_job.erratum.archs:
         return jira_job.erratum.archs
--- a/newa/models/base.py
+++ b/newa/models/base.py
@@ -41,24 +41,24 @@ class Arch(Enum):
                       preset: Optional[list['Arch']] = None,
                       compose: Optional[str] = None) -> list['Arch']:
 
-        _exclude = [Arch.MULTI, Arch.SRPMS, Arch.NOARCH, Arch.I386, Arch.I686, Arch.S390]
-        _all = [Arch(a) for a in Arch.__members__.values() if a not in _exclude]
-        _default = [Arch(a) for a in ['x86_64', 's390x', 'ppc64le', 'aarch64']]
-        _default_rhel7 = [Arch(a) for a in ['x86_64', 's390x', 'ppc64le', 'ppc64']]
+        _valid = [Arch.X86_64, Arch.AARCH64, Arch.S390X, Arch.PPC64LE, Arch.PPC64]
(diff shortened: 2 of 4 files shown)


brozs:
Both points addressed in fb04318:

- Replaced `list(set().intersection(set()))` with order-preserving `[a for a in preset if a in _valid]`
- Replaced `Arch('noarch')` / `Arch('multi')` with `Arch.NOARCH` / `Arch.MULTI`

Note: both patterns were pre-existing in the code, not introduced by this PR — cleaned up while we're here.

brozs:
One trade-off worth flagging on the refactor from exclude-list to explicit whitelist:

**Old approach** (exclude-list): if a new `Arch` enum member is added (e.g. `RISCV64`), it automatically becomes a valid test architecture — no code change needed beyond the enum.

**New approach** (whitelist): a new enum member must be explicitly added to `_valid` to be recognized as a test target. If forgotten, tests silently won't run on that arch.

Both have merit — the whitelist is more readable and makes version-specific overrides cleaner, but the exclude-list is safer against omission when new architectures are introduced.

Happy to revert to an exclude-list approach if you prefer the auto-inclusion behavior. Let me know.

kkaarreell:
Thank you for the change. It looks good on the first sight. But we need to verify if architectures are properly set also for `newa event --compose RHEL-6 ...` event type. 
Either way, let's give it a bit more "manual" testing first.

kkaarreell (approved):
LGTM

PR #387, opened 2026-07-14
Make file conflict checks respect active filters

kkaarreell:
The "files already exist" checks in jira, schedule, and execute commands blocked execution even when active filters (--issue-id-filter, etc.) would match no existing files. Now when filters are active, only matching jobs are considered for the conflict check.

## Summary by Sourcery

Respect active job filters when checking for existing job files in jira, schedule, and execute commands to avoid blocking operations on non-matching jobs.

Bug Fixes:
- Ensure file conflict checks in jira, schedule, and execute commands only consider jobs matching active filters instead of all existing job files.

Enhancements:
- Introduce a filtered file presence helper that delegates to job loaders when filters are active and falls back to a fast filesystem check otherwise.

### DIFF
--- a/newa/cli/commands/jira_cmd.py
+++ b/newa/cli/commands/jira_cmd.py
@@ -20,7 +20,7 @@ from newa.cli.jira_helpers import (
     _parse_issue_mapping,
     _process_issue_config,
     )
-from newa.cli.utils import initialize_state_dir, test_file_presence
+from newa.cli.utils import initialize_state_dir, test_filtered_file_presence
 
 
 @click.command(name='jira')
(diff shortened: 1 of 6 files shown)


kkaarreell:
● Both suggestions are reasonable in principle but provide minimal practical benefit here:
   
  1. Centralizing the filter-active check: The 4-attribute check appears in exactly one place — inside test_filtered_file_presence. It's not actually duplicated.
  Adding a CLIContext property like has_active_action_filters would make that one line slightly more readable, but it's adding an abstraction for a single call site.
  If a new filter type is added in the future, you'd need to update either the property or the function — the maintenance cost is the same. I'd say low benefit.

  2. Tighter typing for load_jobs: A Protocol requiring filter_actions as a keyword would be more precise than Callable[..., Iterator[Any]]. But the function is
  internal, called from exactly 3 places, all passing ctx.load_*_jobs bound methods. A misuse would be caught immediately at runtime. The Protocol adds boilerplate
  (5+ lines) for negligible safety gain in practice. I'd say low benefit.

  Neither suggestion is wrong — they're just premature for a utility with 3 call sites. If the pattern spreads further (more filters, more commands using it), then
  centralizing would start paying off.

PR #392, opened 2026-08-04
Support reportportal: null in recipe dimensions to disable RP per com…

kkaarreell:
…bination

Allow recipe authors to opt specific combinations out of ReportPortal reporting by setting reportportal: null in a dimension or adjustment. The null is sticky — once set, it cannot be overridden back by later merge sources including CLI fixtures, which is a deliberate deviation from the normal priority rule. The null override logic is intentionally limited to the reportportal key to avoid errors with other keys.

## Summary by Sourcery

Allow recipes to explicitly disable ReportPortal per combination using `reportportal: null`, introduce a human-readable `description` attribute for combinations, and ensure CLI and reporting flows respect these settings.

New Features:
- Support setting `reportportal: null` in fixtures, dimensions, or adjustments to opt specific or all combinations out of ReportPortal reporting, with this opt-out treated as sticky against later overrides.
- Add a `description` attribute on recipe combinations and propagate it through requests for use in Jira summaries and as a fallback for ReportPortal suite descriptions.

Enhancements:
- Update ReportPortal launch handling and Jira/report summaries to operate only on RP-enabled jobs and to correctly count and display requests that report to RP.
- Preserve default ReportPortal configuration only when it is not mentioned in fixtures, ensuring explicit `null` is not overridden.
- Extend Jinja2 rendering of request attributes to include the new `description` field while avoiding spurious 'None' string values.

Documentation:
- Document the new `description` attribute semantics and its interaction with ReportPortal and Jira summaries, including Jinja2 support and examples.
- Document how to disable ReportPortal for specific or all combinations via `reportportal: null` in recipes and how this differs from the global CLI `--no-reportportal` option.

Tests:
- Add unit tests and sample recipe YAMLs covering `reportportal: null` behavior, sticky null semantics, merge logic, description merging, and backwards compatibility for recipes without `description`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -940,7 +940,17 @@ In this example, `full` scenario combinations will report to ReportPortal as con
 
 **Note:** Once `reportportal` is set to `null` for a combination, it cannot be overridden back to a non-null value by later merge sources, including CLI-provided options (e.g. `--fixture`). This is a deliberate deviation from the general priority rule where CLI options and fixtures normally take precedence. The rationale is that an explicit `null` represents an intentional opt-out from RP reporting for that combination.
 
-To disable ReportPortal reporting for all combinations at once, use the `schedule --no-reportportal` CLI option instead.
+To disable ReportPortal reporting for all combinations of a recipe, set `reportportal: null` in `fixtures`. This has the same effect as the `schedule --no-reportportal` CLI option but is scoped to the recipe itself:
+
+```yaml
+fixtures:
+    reportportal: null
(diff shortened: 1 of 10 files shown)


jharuda:
LGTM.

kkaarreell:
@sourcery-ai review

jharuda:
Tested on the new version `newa-0.1.dev888-1.20260804130708309425.pr392.fc44.noarch`. Everything seems fine.

